PR #10, opened 2012-12-22
Api disk check

Guimove:
Replaced the disk check template by a true json file.
is it ok ?


Guimove:
Not working sorry

Capture d e cran 22:12:12 20:46

miniche:
It is not in JSON because templates generating is not implemented in Ajax methods yet.
Maybe in the future, but, for the moment, the server will generate full html itself.

Guimove:
Yes I realized when I saw the error

PR #11, opened 2012-12-22
Api route setup

Guimove:
Routes for api are now created.
You can now access the APIs with : 

localhost:8080/api/disks_check
localhost:8080/api/full_check
localhost:8080/api/date_check

You need the "routes" python package : http://routes.readthedocs.org/en/latest/#url_for

Install with : easy_install routes

src : http://tools.cherrypy.org/wiki/RoutesUrlGeneration

All seems working for me

Capture d e cran 22:12:12 23:15


Ninir:
Why do you have 2 commits which are not yours @Guimove?

centralreport/web/server.py
```diff
@@ -71,6 +75,14 @@ def __init__(self):
 
         self.start()
 
+    def setupRoutes(self):
+        d = cherrypy.dispatch.RoutesDispatcher()
```
Ninir: Please don't use such vars, it's hard to know what it really is. Something like "dispatcher" is better IMO (In my opinion).

centralreport/web/server.py
```diff
@@ -71,6 +75,14 @@ def __init__(self):
...
+        d.connect('api_disks_check', '/api/disks_check', controller=Pages(self.env).api_disks_check)
```
Ninir: As you are using check in every routes, I suggest to use something like: `/api/check/disks`

centralreport/web/server.py
```diff
@@ -71,6 +75,14 @@ def __init__(self):
...
+        d.connect('api_date_check', '/api/date_check', controller=Pages(self.env).api_date_check)
+        d.connect('api_full_check', '/api/full_check', controller=Pages(self.env).api_full_check)
+        dispatcher = d
+        return dispatcher
```
Ninir, line 84: It is useless to assign a var with `d` and to return it as you can return d directly

Guimove, line 84: I followed the coding standards of "Routes" 
look ==> http://routes.readthedocs.org/en/latest/

Guimove:
The others commits followed me with the checkout, sorry.
I've made my branch from "develop-doc"

Ninir:
@Guimove I see... well in fact, documentations are sometimes written by engineers who are used to think as senior developpers not as new developpers, leading to a confusion. I am thanking you for your refacto: it will be easier for all of us, (and probably open-source developpers, later ;)) to have meaningful variables and functions rather than ternary operators everywhere and engineers vars ^^

Ninir:
Not needed commits removed

Ninir:
@miniche @Guimove Didn't check using the app, but this seems good

miniche:
- Bash scripts must be updated to install the "route" python package during CentralReport installation ("easy_install" requires setuptools, be sure to do this after installing setuptools on Debian/Ubuntu)
- Javascript needs some changes to work with new routes (ajax updates). I will do it next week :)

miniche:
Good for me

PR #19, opened 2013-01-11
Interval between two checks

miniche:
Interval between two checks will be configurable in the config file!


centralreport/cr/threads.py
```diff
@@ -51,7 +56,7 @@ def run(self):
         Checks.hostEntity = self.MyCollector.get_infos()
 
         while Checks.performChecks:
-            if self.tickCount == self.tickPerformCheck:
+            if self.tickCount >= self.tickPerformCheck:
```
Ninir: As python will swap your check into this one `self.tickPerformCheck <= self.tickCount`, I suggest to make its work easier swapping them ourselves ^^

centralreport/cr/tools.py
```diff
@@ -27,6 +27,9 @@ class Config:
     CR_CONFIG_ENABLE_DEBUG_MODE = False  # False = Production environment. True = debug/test/develop environment.
     CR_HOST_UUID = ''  # Universal Unique IDentifier for the current host. '' = not defined yet.
 
+    # Default interval between two checks (use this if not available in the config file)
+    CR_CONFIG_DEFAULT_CHECKS_INTERVAL=int(60)
```
Ninir: Please add spaces before and after the `=` operator

centralreport/cr/tools.py
```diff
@@ -60,6 +63,7 @@ class Config:
             },
         'Checks':
             {
+                'interval': '60',
```
Ninir: I suggest to alphabetize keys to improve readability

centralreport/web/js/cr.ajax.js
```diff
@@ -57,6 +60,14 @@ var verifyIsNewCheckIsAvailable = function () {
 
     $.ajax('/api/check/date')
         .done(function(data) {
+
+            // Getting interval between two checks
+            if (isNaN(data["checks_interval"])) {
+                checksInterval = CHECKS_INTERVAL_DEFAULT
+            } else {
+                checksInterval = parseInt(data["checks_interval"]);
+            }
```
Ninir:
``` python
checksInterval = isNaN(data["checks_interval"]) ? CHECKS_INTERVAL_DEFAULT : parseInt(data["checks_interval"]);
```

PR #27, opened 2013-02-10
os module, and executeCommand function

miniche:
This function will be use everywhere we need to execute an os command.
It's more flexible, if we want to replace the "subprocess.Popen" call in the future for example. Returns the result string.

Todo: 
- [x] Create the os module
- [x] Design the executeCommand function
- [x] Replace code everywhere is needed to use the new function instead of "subprocess" call

### DIFF
--- a/centralreport/cr/os.py
+++ b/centralreport/cr/os.py
@@ -13,14 +13,14 @@ import subprocess
 
 def WIP_executeCommand(str_command):
     """
-        This function executes an os command. Command can contains pipes, executed as subcommands.
+        This function executes an os command. str_command can contain pipes, executed as subcommands.
         If your command contains vars with a space, use " " to escape words (eg: "Volume Name")
-        If you want to use an double quote (") in your command, escape it with \\ (eg: \\":\\" return ":")
+        If you want to use a double quote (") in your command, escape it with \\ (eg: \\":\\" return ":")
 
         Returns the result of the command (string).
     """
 
-    # Command can contains pipes. At each pipe, we execute a "subcommand".
+    # Command can contain pipes. At each pipe, we execute a "subcommand".
     list_pipes = str_command.split('|')
 
     # Catch all commands result in a list. Used for subcommands stdout/stdin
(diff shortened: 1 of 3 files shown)


centralreport/cr/os.py
```diff
@@ -0,0 +1,44 @@
...
+    https://github.com/miniche/CentralReport/
+"""
+
+import shlex
+import subprocess
+
+
+def WIP_executeCommand(str_command):
+    """
+        This function executes an os command. Command can contains pipes, executed as subcommands.
```
Ninir:
``` python
str_command can contain pipes
```

'Find it weird to have "Command", which is implicitly related to the str_command parameter

centralreport/cr/os.py
```diff
@@ -0,0 +1,44 @@
...
+        If your command contains vars with a space, use " " to escape words (eg: "Volume Name")
+        If you want to use an double quote (") in your command, escape it with \\ (eg: \\":\\" return ":")
```
Ninir: Typo: `an`

centralreport/cr/os.py
```diff
@@ -0,0 +1,44 @@
...
+
+        Returns the result of the command (string).
+    """
+
+    # Command can contains pipes. At each pipe, we execute a "subcommand".
```
Ninir: Typo: s to contain

centralreport/cr/os.py
```diff
@@ -0,0 +1,44 @@
...
+        print list_command
+        # If it's not the first occurence, stdin is the last command executed.
+        if 0 == i:
+            list_results.append(subprocess
+                                .Popen(list_command, stdout=subprocess.PIPE))
+        else:
+            list_results.append(subprocess
+                                .Popen(list_command, stdout=subprocess.PIPE, stdin=list_results[i - 1]
+                                .stdout))
+            # End the previous subprocess (http://docs.python.org/2/library/subprocess.html#replacing-shell-pipeline)
```
Ninir: Typo: Ends

miniche:
Time to review! :innocent:

Guimove:
Seems to be good for me

PR #33, opened 2013-02-16
Refactoring log functions and adding rotating log files

miniche:
TODO:
- [x] Refactor CentralReport log functions
- [x] Add RotatingFileHandler methods to CentralReport logs
- [x] Refactor CentralReport debug mode (if possible)

### DIFF
--- a/bash/log.inc.sh
+++ b/bash/log.inc.sh
@@ -7,8 +7,8 @@
 # https://github.com/miniche/CentralReport/
 # ------------------------------------------------------------
 
-# Logs will be written in this file. Installer or Uninstaller have administrative privileges, we can write in /var/log.
-ERROR_FILE="/var/log/centralreport.log"
+# Logs will be written in this file, it doesn't require administrative privileges.
+ERROR_FILE="/tmp/centralreport_install.log"
 
 # Writes a message in the log file
 function logFile() {
@@ -59,6 +59,13 @@ function printLightBox() {
 
 # Writes a message on the error output and in the log file
 function logError() {
+    logConsole " "
     echo -e "\033[0;31m$*\033[0m" >&2
     logFile "$*"
+}
+
+# Writes a message in the system log (typically in /var/log/system.log)
+# This function doesn't require administrative privileges.
+function logSystem() {
+    logger -t "CentralReport" "$*"
 }
--- a/centralreport/cr/log.py
+++ b/centralreport/cr/log.py
@@ -11,6 +11,40 @@ import logging
 import logging.handlers
 import sys
 
+# Our custom logger object. Initialized on the first use.
+crLogger = None
+enable_debug_mode = False
+
+
+def getCrLogger():
+
+    global crLogger
+    global enable_debug_mode
+
+    if crLogger is None:
+        # Using python rotating log function
+        crLogger = logging.getLogger()
+
+        # In debug mode, log file must be in "/tmp", due to system permissions.
+        logFilename = '/var/log/centralreport.log' if enable_debug_mode is False else '/tmp/centralreport.log'
+
+        crLogger.setLevel(logging.DEBUG)
+        customRotatingFileHandler = logging.handlers.RotatingFileHandler(logFilename, maxBytes=6000, backupCount=5)
+        customRotatingFileHandler.setFormatter(logging.Formatter(fmt='%(levelname)s \t %(asctime)s \t %(message)s',
+                                                                 datefmt='%m/%d/%Y %I:%M:%S'))
+
+        crLogger.addHandler(customRotatingFileHandler)
+
+        if enable_debug_mode:
+            customConsoleHandler = logging.StreamHandler(stream=sys.stdout)
+            customConsoleHandler.setLevel(logging.DEBUG)
+            customConsoleHandler.setFormatter(logging.Formatter(fmt='%(levelname)s \t %(asctime)s \t %(message)s',
+                                                                datefmt='%m/%d/%Y %I:%M:%S'))
+
+            crLogger.addHandler(customConsoleHandler)
+
+    return crLogger
+
 
 def configLog(enable_debug_mode=False):
     """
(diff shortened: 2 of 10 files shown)


miniche:
Ok guys, you can review this PR :grin:

NB: The refactor of debug mode and convention name will be done in next pull request.

Guimove:
Tested and Approved, good luck with merge conflicts

bash/log.inc.sh
```diff
@@ -7,8 +7,8 @@
 # https://github.com/miniche/CentralReport/
 # ------------------------------------------------------------
 
-# Logs will be written in this file. Installer or Uninstaller have administrative privileges, we can write in /var/log.
-ERROR_FILE="/var/log/centralreport.log"
+# Logs will be written in this file, it doesn't require administrative privileges.
+ERROR_FILE="/tmp/centralreport_install.log"
```
Ninir, line 11: I would add it in `/tmp/centralreport/`

centralreport/cr/log.py
```diff
@@ -2,50 +2,93 @@
...
+        cr_logger.setLevel(log_level)
+
+        # In debug mode, log file must be in "/tmp", due to system permissions.
+        log_filename = '/var/log/centralreport.log' if debug_mode_enabled is False else '/tmp/centralreport.log'
 
-        # Writing only "INFO" or more important messages in a log file (production environement)
-        logging.basicConfig(filename='/var/log/centralreport.log',
-                            format='%(levelname)s \t %(asctime)s \t %(message)s',
-                            level=logging.INFO, datefmt='%m/%d/%Y %I:%M:%S')
+        # Max size per log file: 5 MB (1024 * 1024 * 5). Number of log files will be kept for archive: 2.
```
Ninir: It seems you have something missing in your second sentence:
`Number of log files "that" will be kept for archive: 2.` ?

Btw, I would write: `2 files will be kept as archive.`

centralreport/cr/log.py
```diff
@@ -2,50 +2,93 @@
...
+
+            cr_logger.addHandler(custom_console_handler)
+
+    return cr_logger
 
 
 def writeDebug(text):
-    logging.debug(text)
+    """
+        Writes a debug message. Only useful for testing and debugging purposes.
```
Ninir: You could use it on your own, meaning that saying "only" is a bit restrictive.
Why not saying "Useful for testing and debugging purposes."?

centralreport/cr/log.py
```diff
@@ -2,50 +2,93 @@
...
+    """
+        Writes a standard message. Will be written in the current log file.
+    """
+    get_cr_logger().info(text)
 
 
 def writeError(text):
-    logging.error(text)
+    """
+        Writes an error message: an anormal situation, but non-critical.
```
Ninir: Typo: `aBnormal`.

`Writes an error message for non-critical abnormal situations.`

centralreport/cr/log.py
```diff
@@ -2,50 +2,93 @@
...
+        Will be written in the current log file.
+    """
+    get_cr_logger().error(text)
 
 
 def writeCritical(text):
-    logging.critical(text)
+    """
+        Writes a critical error. This error requires to stop the current action.
```
Ninir: `Writes a critical error stopping the current action.`

centralreport/cr/log.py
```diff
@@ -2,50 +2,93 @@
...
+        Will be written in the current log file.
+    """
+    get_cr_logger().critical(text)
 
 
 def writeFatal(text):
-    logging.fatal(text)
+    """
+        Writes a fatal error. This error requires to stop current application.
```
Ninir: `Writes a fatal error stopping the application.`

Ninir:
Excepted comments, seems ok. You'll have to merge conflicts... enjoy :D

bash/log.inc.sh
miniche, line 11: I'm not sure if it's very useful. This file is temporary, and it used only if an error occurs during install/uninstall.
And this file will be deleted immediately after a restart.

But, you give me an idea: I think we can delete this file if no error occured during installation, no?

centralreport/cr/log.py
miniche: Yes, it's restrictive. In fact, debug message are ignored by default, not written in the log file neither on standard output.
When you launch CR with the "test.py" script, "debug mode" is enable, and these messages will be kept (log file + stdout). So, for me, "only" is the good work. I will add a clearer explanation in the comment :)

centralreport/cr/log.py
miniche: You change the meaning of my sentence. This function doesn't stop the current action.
By "this error requires to stop the current action.", I just want to define a critical error.

centralreport/cr/log.py
miniche: Same answer as the previous comment.

centralreport/cr/log.py
Ninir: Your second sentence is not crystal-clear. I know that, at the time I was writing my comment that without looking at the code, and  seeing at the doc that your function will write the log. If we translate your sentence: "Cette erreur force à stopper l'action courante."
IMO, your second sentence is simply useless as it brings no value to the function. It is only an information for you, but as a developer... I don't really care :/

centralreport/cr/log.py
Ninir: Same answer as my previous comment :)

bash/log.inc.sh
Ninir, line 11: Well, creating the folder IS useful as it will contains our log files. I think our structure should be as modular as possible, and as silent as possible, meaning we should not polluate /tmp directly, putting everything in a folder we control.

The purpose of /tmp is to be temporary, so adding a temporary folder isn't a real matter, is it? :)

centralreport/cr/log.py
Ninir: If debug messages are ignored, then "only" is well chosen, otherwise it should not be there. Thanks for sharing :)

centralreport/cr/log.py
miniche: Ok, I understand what you want to say.
For me, it's important to define what is a critical, a fatal error, or a "simple" error, and when these functions must be used (and when we shouldn't use them). 
I will try to find a better sentence.

centralreport/cr/log.py
miniche: Same answer as my previous comment. +3

bash/log.inc.sh
miniche, line 11: But we have only one file to write into /tmp: the install log.
All other logs generated by the main application go into /var/log. It isn't useless to create a folder for only tmp file?

Ninir, line 11: In fact, after thinking of it again, a dir seems useless here yes... this is a `/tmp` dir, not the `/var/log` as you said.
Forget about this

centralreport/cr/log.py
```diff
@@ -2,51 +2,94 @@
...
+        Contains log functions to work with native Python logging modules
 
     https://github.com/miniche/CentralReport/
 """
 
 import logging
+import logging.handlers
 import sys
 
+# Our custom logger object. Initialized on the first use.
```
Ninir, line 14: As you point out the fact that initialization will be done on first use in the `get_cr_logger` description, your second sentence seems useless or, at least, repetitive.

centralreport/cr/log.py
```diff
@@ -2,51 +2,94 @@
...
-        logging.basicConfig(filename='/var/log/centralreport.log',
-                            format='%(levelname)s \t %(asctime)s \t %(message)s',
-                            level=logging.INFO, datefmt='%m/%d/%Y %I:%M:%S')
+        # Max size per log file: 5 MB (1024 * 1024 * 5). 2 files will be kept as archive.
+        custom_rotating_fileHandler = logging.handlers.RotatingFileHandler(log_filename,
+                                                                           maxBytes=5242880,
+                                                                           backupCount=2)
 
-    else:
+        custom_rotating_fileHandler.setFormatter(logging.Formatter(fmt='%(levelname)s \t %(asctime)s \t %(message)s',
```
Ninir: What do you think of this format?

`[01:01:2000 01:01:01] - LEVELNAME: Message`

centralreport/cr/log.py
Ninir: For instance, this is taken from the Monolog library, which is now a reference in the PHP world:

``` php
/**
* Adds a log record at the DEBUG level.
*
* @param  string  $message The log message
* @param  array   $context The log context
* @return Boolean Whether the record has been processed
*/
public function addDebug($message, array $context = array())
{
    return $this->addRecord(self::DEBUG, $message, $context);
}
```

Ninir: Even better: https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php

Ninir:
Here is a proposal for the logger object:

Logger Object: Here to log things at a given log level.

User functions:
- log_debug:

``` python
def log_debug(text):
    """
        Adds a log record at the DEBUG level.
    """
    get_cr_logger().debug(text)
```
- log_info:

``` python
def log_info(text):
    """
        Adds a log record at the INFO level.
    """
    get_cr_logger().info(text)
```
- log_error:

``` python
def log_error(text):
    """
        Adds a log record at the ERROR level.
    """
    get_cr_logger().error(text)
```
- log_crit:

``` python
def log_crit(text):
    """
        Adds a log record at the CRITICAL level.
    """
    get_cr_logger().critical(text)
```
- log_fatal:

``` python
def log_fatal(text):
    """
        Adds a log record at the FATAL level.
    """
    get_cr_logger().fatal(text)
```

centralreport/cr/log.py
miniche: [01:01:2000 01:01:01] - LEVELNAME \t Message

bash/log.inc.sh
miniche, line 11: Oh! We agree together! Wonderful! :smile:

PR #38, opened 2013-02-22
Code refactor using PEP8

Guimove:
- [x] Remove CamelCase (methode and variables)
- [x] Remove "as" in import
- [x] Clean

### DIFF
--- a/centralreport/centralreport.py
+++ b/centralreport/centralreport.py
@@ -13,17 +13,17 @@ import sys
 import time
 import os
 
-import cr.log as crLog
-import cr.threads as crThreads
-import cr.utils.text as crUtilsText
+from cr import log
+from cr import threads
+from cr.utils import text
 from cr.daemon import Daemon
 from cr.tools import Config
 from web.server import WebServer
 
 
 class CentralReport(Daemon):
-    isRunning = True  # Deamon status
-    startingDate = None
+    is_running = True  # Deamon status
+    starting_date = None
 
     # Threads
     checks_thread = None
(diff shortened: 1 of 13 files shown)


miniche:
I think it's better to keep "as" import alias. It's more readable, and we can identify immediately if the module is a "python module" or "cr module".
I'll look tonight.

miniche:
So, after a long meditation...
It's very difficult to keep the code simple in Python coming from other languages. Python follow the KISS principle, and our code is better without "as" aliases. 

In fact, the main philosophy of custom modules is extend Python possibilities. We must follow the main recommendations and the operating principle of all Python modules, to keep consistency and to be more readable by other developers. Sometime, it's very hard to do simple things...

Ninir:
@miniche I strongly disagree with the fact that we "MUST" follow the main recommandations. The PEP 8 style guide is well... a recommandation, not a "must-be-followed guide". You have different style guides in this world, for every languages, meaning you can find other that better fit your needs.

For example, there is also the Google Python style guide, which you could find [here](http://google-styleguide.googlecode.com/svn/trunk/pyguide.html).

The way you write YOUR code is the way YOU find it better. This means you CAN follow recommandations but you're not force to follow everything in it you know :)

miniche:
@Ninir Just a clarification.

All right, we need to use tools and methods that seem to be the best. But, I think that it's important to say that you must follow the MAIN recommandations for several reasons:
- PEP has been specially written to correspond to Python philosophy
- All major projects and developers follow these recommandations.

Yeah, I'm the author of my code, and I still have some freedoms to design my code how I wish. But we are several persons working on CentralReport and it's important to keep uniformity.
We make this code available for all the community, and especially for all Python developers. I could write all my code following the Java style recommandations. At this moment, CentralReport should become very different of what happens in every other Python projects, without consistency: for every person who wants to collaborate on our project, it will be harder to read and follow custom rules instead of general rules.

The world must have some rules. Alright, it's just a recommandation, but we must follow some rules to keep consistency in our project, but also with the world, with other Python projects.

The Google style guide follow the PEP 8, it's juste more synthetic and clearer.
And, I think that we can't publish our custom recommandations in Python for the moment :stuck_out_tongue_winking_eye:

Finally, I just only want to add a nuance. I specify we must follow the "main" recommandations of the PEP, not the entire PEP. Some rules are so old, like the line of 80 characters at maximum....
But, except this, I think it's better to follow these recommandation instead of doing what we want.

Ninir:
@miniche AS I said:

> This means you CAN follow recommandations but you're not force to follow everything in it you know :)

Meaning if you want to use aliases (`as`), you can. I agree, we have to follow rules in order for other developpers not to be lost, but if you find that using this is needed, or at least you find it better, do it.

I'm not saying we should not follow styleguides but adapt what we think is better for the solution. Nothing less, nothing more.

miniche:
Basically, I think we agree together. It's just a question of vocabulary.

I think it's better to say "We must follow recommandations" instead of "We can follow recommandations".
If we "just" can follow recommandations, we will do what we want and we only observe recommandations when the sun is up..

If we say "must", we will consider such recommandations are mandatory, unless they are stupid or impractical in CentralReport. If import aliases help to be more clearer, we can implement them but only when we don't have an other choose.

PR #39, opened 2013-02-22
Updated install of dev tools

Guimove:


tools/install_dev_tools.sh
```diff
@@ -12,6 +12,7 @@
 
 source ../bash/vars.sh
 source ../bash/functions.inc.sh
+source ../bash/log.inc.sh
```
miniche, line 14: Logical order:
- vars
- log
- functions

tools/install_dev_tools.sh
```diff
@@ -11,6 +11,7 @@
 # It's be very helpfull on a dev host.
 
 source ../bash/vars.sh
+source ../bash/log.inc.sh
 source ../bash/functions.inc.sh
 
 echo "------------------------------------------"
```
miniche: Maybe you can use the "lightbox" function (see in log.inc.sh).

miniche:
Can be cool: Use a global error var instead of "return ${ERROR_CODE}" and display a red lightbox when the install fail.
You can test before each "displayAndExec" if an error has already occurred.

tools/install_dev_tools.sh
```diff
@@ -11,62 +11,46 @@
...
+source ../bash/log.inc.sh
 source ../bash/functions.inc.sh
 
-echo "------------------------------------------"
-echo "CentralReport dev tools installer"
-echo "------------------------------------------"
+#=============================================================================
+# List of library to install
+# Can be updated as your needs
+declare -a LISTE='([0]="CherrezyPy" [1]="Jinja2" [2]="Routes")'
```
miniche: CherrezyPy?

tools/install_dev_tools.sh
```diff
@@ -11,62 +11,46 @@
...
 
 source ../bash/vars.sh
+source ../bash/log.inc.sh
 source ../bash/functions.inc.sh
 
-echo "------------------------------------------"
-echo "CentralReport dev tools installer"
-echo "------------------------------------------"
+#=============================================================================
+# List of library to install
```
miniche: `libraries`, no ?

Guimove:
Here is how an error looks : 

Capture_d e cran_27_02_13_14_09-2

tools/install_dev_tools.sh
```diff
@@ -11,62 +11,46 @@
...
-sudo rm -Rf ${JINJA_DIR}
-
-echo "Jinja2 is installed!"
-
+for i in ${!LISTE[*]}
+do
+    displayAndExec "Installing ${LISTE[${i}]}..." sudo easy_install ${LISTE[${i}]}
+    RETURN_CODE="$?"
+    if [ ${RETURN_CODE} -ne 0 ]; then
+        printLightBox red "Woops... Something went wrong with the installation of ${LISTE[${i}]}"
```
Ninir: `Woops... Something went wrong installing ${LISTE[${i}]}` seems shorter :D (poke @miniche)

miniche:
In the installer and the uninstaller, lightbox looks like this:

Capture d e cran 2013-02-27 a 14 38 59
- In an error lightbox, we have to empty red light before and after the main text,
- In each line in a lightbox, we start with a space to be clearer,
- We have a blank line before and after lightbox,
- I think that you can remove dashes in the header, or you can add more dashes to do a full line.

Ninir:
Why not making a `printLightboxBlock` with 2 parameters: the color and the text.
Thus, you could simply reduce the amount of lines displayed. This seems not maintainable as you have to add a newline before and after the text which is, well, repetitive. This function could also cut the sentence to fit in 80 characters if it's too long. Something like:

``` bash
printLightBox red " "
printLightBox red "Woops... Something went wrong installing ${LISTE[${i}]}"
printLightBox red "Read the log file in ${ERROR_FILE} for further informations"
printLightBox red " "
```

to:

``` bash
printLightBoxBlock red "Woops... Something went wrong installing ${LISTE[${i}]}. Read the log file in ${ERROR_FILE} for further informations"
```

What do you think guys?

Guimove:
Ok I'll make this in the future. We have to think about the best solution. 
Issue #42

tools/install_dev_tools.sh
```diff
@@ -11,62 +11,119 @@
...
+
+    printTitle "Installing Libraries..."
+
+    for i in ${!LISTE[*]}
+    do
+        displayAndExec "Installing ${LISTE[${i}]}..." sudo easy_install ${LISTE[${i}]}
+        RETURN_CODE="$?"
+        if [ ${RETURN_CODE} -ne 0 ]; then
+            printLightBox red " "
+            printLightBox red "Woops... Something went wrong installing ${LISTE[${i}]}"
```
miniche: Can you add a space before "Woops"?
I think that it's more readable in the console.

(Idem for all red lightboxes)

tools/install_dev_tools.sh
```diff
@@ -11,62 +11,119 @@
...
-echo "------------------------------------------"
-echo "CentralReport dev tools installer"
-echo "------------------------------------------"
+#=============================================================================
+# List of libraries to install
+# Can be updated as your needs
+declare -a LISTE='([0]="CherryPy" [1]="Jinja2" [2]="Routes")'
+#=============================================================================
+
+printLightBox blue "---------------------- CentralReport dev tools installer -----------------------"
```
miniche, line 24: What do you think about adding a "clear" command before the header ?

miniche:
Seems good for me!

PR #41, opened 2013-02-23
Fixed logging.StreamHandler for Python 2.6

miniche:
Between Python 2.6 and 2.7, the StreamHandler parameters has been updated.

In 2.7: http://docs.python.org/2/library/logging.handlers.html#logging.StreamHandler
And in 2.6: http://docs.python.org/2.6/library/logging.html#logging.StreamHandler

### DIFF
--- a/centralreport/cr/log.py
+++ b/centralreport/cr/log.py
@@ -45,7 +45,7 @@ def get_cr_logger():
 
         if debug_mode_enabled:
             # In debug mode, we display logs on the standard output too.
-            custom_console_handler = logging.StreamHandler(stream=sys.stdout)
+            custom_console_handler = logging.StreamHandler(sys.stdout)
             custom_console_handler.setLevel(logging.DEBUG)
             custom_console_handler.setFormatter(_formatLogger())
 


Ninir:
When you provide such changes, please add sources :)

Ninir:
Btw, aren't we using the Python 2.7 version? If not then.. why not?

miniche:
CentralReport works at least with Python 2.6 and 2.7 (not tested on 2.3, 2.4 and 2.5 yet).
We must support all Python 2.x versions as much as possible.

And, what sources? No external source for this, it's just from tests and official Python documentation.
If you want, I can add a link to Python doc at each line... :smile:

Ninir:
@miniche By source, I indeed mean python doc, meaning you can add them in the Pull Request description, not on each line of the file, which is as useless as possible.

miniche:
Oh... Yeah. I didn't understand. Sorry.
So... Well... Difficult to say but... It seems... Wait... Yes. You're right. :pensive:

Ninir:
@miniche :heart: hehe :P 

It's just easier for dumb people like me to understand :P

miniche:
Ready to merge?

Ninir:
Seems good

Issue #42, opened 2013-02-27
Create some functions to reduce code size


Guimove:
We should think about the creation of some functions to reduce code size.

Something like this :

``` bash
printLightBox red " "
printLightBox red "Woops... Something went wrong installing ${LISTE[${i}]}"
printLightBox red "Read the log file in ${ERROR_FILE} for further informations"
printLightBox red " "
```

to:

``` bash
printLightBoxBlock red "Woops... Something went wrong installing ${LISTE[${i}]}. Read the log file in ${ERROR_FILE} for further informations"
```
#39

PR #45, opened 2013-02-28
Creation of the printBox function

Guimove:
Refer to issue #42

### DIFF
--- a/bash/log.inc.sh
+++ b/bash/log.inc.sh
@@ -58,16 +58,16 @@ function printLightBox() {
 }
 
 function trim() {
-    local var=$1
-    var="${var#"${var%%[![:space:]]*}"}"   # remove leading whitespace characters
-    var="${var%"${var##*[![:space:]]}"}"   # remove trailing whitespace characters
-    echo -n "$var"
+    local TEXT="$1"
+    TEXT="${TEXT#"${TEXT%%[![:space:]]*}"}"   # remove leading whitespace characters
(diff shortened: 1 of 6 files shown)


bash/log.inc.sh
```diff
@@ -57,6 +57,34 @@ function printLightBox() {
 
 }
 
+function trim() {
```
miniche, line 68: Can you add a small docstring before the function?
(Like others functions in log.inc.sh or functions.inc.sh, or more detailed if you have other ideas)

(Idem for printBox())

bash/log.inc.sh
```diff
@@ -57,6 +57,34 @@ function printLightBox() {
...
+    local TEXT="$1"
+    TEXT="${TEXT#"${TEXT%%[![:space:]]*}"}"   # remove leading whitespace characters
```
Ninir: Is it a recommandation to add the comment 3 spaces after an instruction?

Ninir:
Seems good looking at the code. :+1:

bash/log.inc.sh
miniche, line 68: What do you think about this model of docstring?

``` bash
#
# Explain the function utilty here
#
# PARAMETERS:
#     .....
# RETURN:
#     ....
#
```

An example:

``` bash
#
# Displays a message on the standard output
#
# PARAMETERS: 
#     $1 = message to display
# RETURN:
#     0 = Success
#     The error code otherwise
#
function printLog {
...
}
```

miniche:
You can merge now. Go go go!

PR #48, opened 2013-03-01
Unification of checks

Guimove:
- [x] Replace some bash function by their python brother 
- [x] Unification of some checks between Linux en Mac

As much as possible !

### DIFF
--- a/centralreport/cr/collectors.py
+++ b/centralreport/cr/collectors.py
@@ -19,6 +19,7 @@ import re
 import socket
 import time
 import os
+import sys
 
 from cr.entities import checks
 from cr.entities import host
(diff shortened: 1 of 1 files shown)


centralreport/cr/collectors.py
```diff
@@ -132,65 +166,27 @@ def get_memory(self):
...
         """
-            Gets the host load average.
+            Gets load average.
         """
 
-        dict_iostat = self.get_IO_stats()
+        loadavg_result = os.getloadavg()
+
+        # On va spliter en fonction des espaces
+        dict_loadavg = loadavg_result
```
miniche: I think this line is not really useful, no?

miniche:
Another possible improvement, in `Debian Collector > get_infos`, lines 280 to 291:

I think you can remplace this (my? :grin:) spaghetti code with only the following command:

``` bash
cat /etc/lsb-release | grep "DISTRIB_RELEASE" | awk 'BEGIN { FS="=" } END { print $2; }'
```

Yeah, I know, you love this command :kissing_heart:
Tested with success on Ubuntu 11.10.

Guimove:
This only work on Ubuntu, /etc/lsb-release doesn't exist on Debian.
But I'll use 

```
platform.linux_distribution()
```

This returns : 

```
('debian', '6.0.7', '')
```

miniche:
Yeah, I know my solution only works on Ubuntu, but the test for Debian has already been done 5 lines before.

And, very good idea for your improvement! I'm impressed :stuck_out_tongue_winking_eye:

miniche:
Seems good for me :+1:

Issue #56, opened 2013-03-28
Serializer Component


Ninir:
In order to reduce code redundancies when serializing data, we should implement a serializer component, which will be in charge to take a data, serializes it in a given format, and returns the serialized object.

IMO, it is way better to do such think in order to get rid of all the serialize/json_serialize methods we could have in each classes. Something like:

``` python
class Serializer():

    def serialize(self, content, format=None):
        """
            Serializes data in a given format.
        """

        # Do some stuff...

        # original_object = content
        # serialized_data = json_dumps(original_object)

        return serialized_data

    def deserialize(self, content, format=None):
        """
            Deserializes a serialized content back to its original type.
        """

        # Do some stuff...

        # serialized_data = content
        # original_object = dict()
        # from StringIO import StringIO
        # io = StringIO('["streaming API"]')
        # original_object = json.load(io)

        return original_object
```

This is an example and should not be taken as is. We could also use reflection which will allow more benefits from it.

miniche:
Very good idea!
It's a work for super @Guimove! :+1:

PR #57, opened 2013-03-29
Creation of the serializer (convert an object in whatever you want)

Guimove:
Issue #56 

In thread.py : 

``` python
 myHost = webservices.Full()
 myHost.host = Checks.host_infos
 myHost.checks.append(check_entity)

 log.log_debug(objects.object_converter(myHost, "json"))
```

Return : 

``` json
{
   "host":{
      "cpu_count":8,
      "uuid":"toto",
      "language":"Python",
      "kernel_version":"12.3.0",
      "hostname":"Dell-Pc",
      "cpu_model":"Intel(R) Core(TM) i7-2720QM CPU @ 2.20GHz\n",
      "os_version":"10.8.3\n",
      "os_name":"Mac OS X\n",
      "kernel_name":"Darwin",
      "architecture":"x86_64",
      "key":"toto",
      "date":"2013-03-29 15:48:45.698042",
      "model":"MacBookPro8,2\n",
      "os":"Mac OS X"
   },
   "checks":[
      {
         "date":"2013-03-29 15:48:47.001335",
         "load":{
            "last1m":"1.70",
            "last15m":"1.64",
            "uptime":84463,
            "last5m":"1.65"
         },
         "disks":{
            "disks":[
               {
                  "used":87758880768.0,
                  "name":"Mac\n",
                  "date":"2013-03-29 15:48:46.799442",
                  "size":119174365184.0,
                  "free":31153340416.0,
                  "unix_name":"/dev/disk1s2"
               },
               {
                  "used":290056925184.0,
                  "name":"Mac DATA\n",
                  "date":"2013-03-29 15:48:47.001213",
                  "size":749812400128.0,
                  "free":459755474944.0,
                  "unix_name":"/dev/disk0s2"
               }
            ]
         },
         "cpu":{
            "idle":"90",
            "system":"4",
            "user":"6"
         },
         "memory":{
            "swap_used":0.0,
            "swap_free":0,
            "resident":1376833536.0,
            "free":125407232.0,
            "inactive":3424808960.0,
            "swap_size":8579194880.0,
            "active":3652145152.0,
            "total":8579194880.0
         }
      }
   ]
}
```


centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+import json
+
+
+def object_converter(data, encod):
+    """
+        Represent instance of a class converted in what you want.
```
Ninir: `Represents an instance of a class converted to a given format.`

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
+# -*- coding: utf-8 -*-
+import json
+
+
+def object_converter(data, encod):
```
Ninir: `format` is better than `encod` IMO

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+        Represent instance of a class converted in what you want.
+        @param encod: Encoded format like JSON...
+        @param data: An object
+        @Return: String that reprent encoded object.
```
Ninir: `reprent` -> `represents`

String that represents the encoded object.

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def serialize(data):
+        """
+            Serialize the object whatever is type
```
Ninir: - `Serializes`
- `...whatever is type is`

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
...
+            return [serialize(item) for item in data]
+        elif isinstance(data, tuple):
+            return tuple(serialize([item for item in data]))
+        elif hasattr(data, '__dict__'):
+            return serialize(data.__dict__)
+        else:
+            return str(data)  # convert to string
+
+    # We can add "if" for all type we want
+    if "json" == encod:
```
Ninir: Is `if encod is "json"` working?

centralreport/cr/utils/object.py
Guimove: whatever his type is

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
...
+                data[key] = serialize(data[key])
+            return data
+        elif isinstance(data, list):
+            return [serialize(item) for item in data]
+        elif isinstance(data, tuple):
+            return tuple(serialize([item for item in data]))
+        elif hasattr(data, '__dict__'):
+            return serialize(data.__dict__)
+        else:
+            return str(data)  # convert to string
```
Ninir: The serialize function could be refactor as is:

``` python

    if isinstance(data, (bool, int, long, float, basestring)):
        return data

    if isinstance(data, dict):
        data = data.copy()
        for key in data:
            data[key] = serialize(data[key])
        return data

    if isinstance(data, list):
        return [serialize(item) for item in data]

    if isinstance(data, tuple):
        return tuple(serialize([item for item in data]))

    if hasattr(data, '__dict__'):
        return serialize(data.__dict__)

    return str(data)  # converts to string
```

centralreport/cr/utils/object.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # We can add "if" for all type we want
+    if "json" == encod:
+        return json.dumps(serialize(data))
+    else:
+        return serialize(data)
```
Ninir: These 4 lines could be refactor to something like this:

``` python
if "json" == encod:
    return json.dumps(serialize(data))

return serialize(data)
```

centralreport/cr/utils/object.py
Ninir: `whatever ITS type` is ;)

centralreport/cr/utils/object.py
miniche: The `is` key is used in Python to test the identity, not the equality. If `encod` is a string, your test is always true.
(RTFM: http://docs.python.org/2/library/stdtypes.html)

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+        Represents an instance of a class converted to a given format.
+        @param encod: Encoded format like JSON...
+        @param data: An object
+        @Return: String that represents encoded object.
+    """
+
+    def serialize(data):
+        """
+            Serializes the object whatever his type is
```
Ninir: its type*

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+import json
+
+
+def object_converter(data, encod):
+    """
+        Represents an instance of a class converted to a given format.
+        @param encod: Encoded format like JSON...
+        @param data: An object
+        @Return: String that represents encoded object.
```
Ninir: `@Return: A string representing the encoded object.`

miniche:
You have just forgot the "module header". You can find it in any other module, e.g.:

``` python
"""
    CentralReport - system module
        Contains functions to interact with the operating system

    https://github.com/miniche/CentralReport/
"""
```

Guimove:
Can be merged ?

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: utf-8 -*-
+
+"""
+    CentralReport - system module
```
Ninir: Well.. not really :smile:

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,51 @@
...
+        Contains functions to interact with the operating system
```
Ninir: Still not the case :smile:

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+
+import json
+
+
+def object_converter(data, encod):
+    """
+        Represents an instance of a class converted to a given format.
+        @param encod: Encoded format like JSON...
+        @param data: An object
```
Ninir: I would switch these 2 params in order to have them in order they are provided into the class signature

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,51 @@
...
+    https://github.com/CentralReport/
+"""
+
+import json
+
+
+def object_converter(data, encod):
+    """
+        Represents an instance of a class converted to a given format.
+        @param encod: Encoded format like JSON...
```
Ninir: Well, this sentence is not friendly to my ears :/
I would say something like `Format to encode the object to`, or something in the idea. What do you think?

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,51 @@
...
+        @param data: An object
+        @Return: A string representing the encoded object.
+    """
+
+    def serialize(data):
+        """
+            Serializes the object whatever its type is
+            @param data: An object
```
Ninir: It seems you forgot the return annotation

Ninir:
Seems good :) :clap: 
Could you remove the `[WIP]` tag from the Pull Request please?

centralreport/cr/utils/objects.py
Guimove: I think I was drunk... :joy:

Ninir: Again? please stop drinking :P

Ninir:
I would renamed `objects.py` to `serializer.py` as the description is exposing a text for a serializer module.
Am I right?

miniche:
Right. At the beginning, I've planned to use this new module with only functions on objects.
In this package (utils), we have some modules distributed by content type (like "text" or "date").

@Guimove have created a transversal function: it can be used for object, but also list, dictionaries, tuple...
We have two chooses:
- Rename this module to `serializer` and rename the function
- Keep this module name, but update the description.

IMO, it can be better to rename this module as this function can serialize more than objects.

Ninir:
Agreed

Guimove:
Ok, so how should I rename this file ?

miniche:
`serializer`, or any think else if you have better idea :)
I think the function "object_converter" must be renamed too.

centralreport/cr/utils/objects.py
```diff
@@ -0,0 +1,52 @@
...
+    CentralReport - serializer module
+        Contains functions to parse data to a serialized format
+
+    https://github.com/CentralReport/
+"""
+
+import json
+
+
+def object_converter(data, encod):
```
Ninir: Btw, what is the purpose of having this function and the serialize one below?

Why not doing just a `serialize` function?

Guimove: Because this two functions are two independant functions who works together as a recusive funtion

Ninir: `serialize` is dependent on `object_converter` for what I'm seeing.

'Got what you want to achieve. Then I would rename `object_converter` to serializer, as it is its purpose

Ninir: > serialize is dependent on object_converter for what I'm seeing.

However, Python is still a bit unknown to me, so maybe these 2 are "independant" in the way they are designed, but looking at the code, they seem dependant.

miniche: Yes, these two functions are highly dependents.
If we want to name this module "serializer", it isn't mandatory to call these functions "serialize" or "object_converter" because we must use the module name to call them.

Just a suggestion, can be improved:
- Call the module `serializer`
- The `object_converter` could be renamed as `encode` (serializer.encode())
- The `serialize` function can be detached from the object_converter and could be renamed as `dump` (serializer.dump(), like in json and pickle python modules)

Ninir: IMO, the name of the module has to be `serializer` and methods should be `serialize`/`unserialize`.

Methods are what a module can do, which is pretty obvious here. Having a synonym has no sense when you could you its main definition.

miniche: Python have not consistency in this case. Json and pickle modules follow my idea, but a lot of other module follow your suggestion...
I prefer your solution, definitely.

Some other examples in the Python community:
https://github.com/mitsuhiko/jinja2/blob/master/jinja2/optimizer.py
https://bitbucket.org/cherrypy/cherrypy/src/799f39c23c8d9982f3b0523a9234f518ec3407ae/cherrypy/lib/encoding.py?at=default

(They have not the same objectives, just for naming examples)

Guimove: I say that they are independent because they can work as separated function.
These two functions have their own goals.

miniche: According by the indentation, the "serialize" function is nested in the "object_convert" function.
Are your sure that `serialize` is available from outside of the object_convert?

Guimove: In this case, they have to work together, but in other situations, you can have 

``` Python
def serialize(data):
    """
        Serializes the object whatever its type is
        @param data: An object
        @Return: A string representing the serialized object.
    """
    if isinstance(data, (bool, int, long, float, basestring)):
        return data

    if isinstance(data, dict):
        data = data.copy()
        for key in data:
            data[key] = serialize(data[key])
        return data

    if isinstance(data, list):
        return [serialize(item) for item in data]

    if isinstance(data, tuple):
        return tuple(serialize([item for item in data]))

    if hasattr(data, '__dict__'):
        return serialize(data.__dict__)

    return str(data)  # converts to string

def object_converter(data, encod):
    """
        Represents a data converted into a given serialized format.
        @param data: An object
        @param encod: Encoded format like JSON...
        @Return: A string representing the serialized object.
    """
    # We can add "if" for all type we want
    if "json" == encod:
        return json.dumps(serialize(data))

    return serialize(data)
```

Do you understand what I meen ?

miniche: For me, it's better: we can use the two functions independently.
After, for the name of functions... Perhaps @Ninir have a good suggestion about it :P

Ninir: Agreed. One class/module should have its Single Responsibility. We'll try to see if we can improve this serializer later.

miniche:
This work is mandatory for the webservices.
@Guimove, will you let me update your module with this https://github.com/CentralReport/CentralReport/pull/57#discussion_r4060657?

I can do this tonight and we could continue to work on the webservices this week.

Guimove:
Are you OK with this commit ?
If yes you can merge.

miniche:
Great! For me, it's ok.

Can you just merge develop in this branch in order to have a Travis build succeeded? :)

centralreport/cr/utils/serializer.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+
+"""
+    CentralReport - serializer module
+        Contains functions to parse data to a serialized format
+
+    https://github.com/CentralReport/
```
Ninir, line 7: Shouldn't the URL point to the project more than the organization?

centralreport/cr/utils/serializer.py
```diff
@@ -0,0 +1,53 @@
...
+
+def object_converter(data, encod):
+    """
+        Represents a data converted into a given serialized format.
+        @param data: An object
+        @param encod: Encoded format like JSON...
+        @Return: A string representing the serialized object.
+    """
+
+    # We can add "if" for all type we want
```
Ninir: Finally, `every` sounds better than all (and i'm pretty sure that it's grammatically more correct)

Guimove: Yep I agree with that.

centralreport/cr/utils/serializer.py
Guimove, line 7: Yes I forgot that

Guimove:
@miniche I you want, you can make those little change and merge this branch if you want.

I don't have my computer this week.

centralreport/cr/utils/serializer.py
miniche, line 7: All module headers have this URL, since the pull request #60.

miniche:
My last commit merge the "object_convert" function into the "serialize" function to keep it simple.
I've added to constants (FORMAT_STRING and FORMAT_JSON) in the module to be consistant in the function arguments.

Waiting for your approval :)

miniche:
I need this feature in the "webservices" branch. Will be merged in the next seconds :-)

PR #52, opened 2013-03-22
Improvements - Was "Communication with CentralReport Online"

Guimove:
## Update 03/03/2014

CentralReport Online is not planned anymore. Another project will be born in few days and will be announced on the [CR blog](http://blog.centralreport.net).
As this pull request adds useful features and many improvements in the core of this project, this branch has been merged into develop.

---
### Original pull request:

This pull request adds the communication with CentralReport Online (centralreport.net).
CentralReport Online is an online tool to follow all your hosts, in a single place. This part is currently in development and will be enhanced in the next weeks.
- [x] Add "requests" library
- [x] Create the "webservice" module
- [x] Create webservices entities
- [x] Clean and code optimization 

Remains to be done:
- [x] Clean values from the collector ('\n' in some variables, or too many spaces in others)
- [x] Clean useless entities ("Full", "Registration")
- [x] Compact the JSON output (on a single line, without useless spaces)
- [x] Cast dates and float to long (like disk spaces or memory data)
- [x] Update the thread "Checks" to send data to the Online part
- [x] Do more tests with the JSON generation
- [x] Add the "Agent name" and "Agent version" in the host json.

_Updated tasks (miniche, 02/04/2013)_
_Updated tasks (miniche, 29/05/2013)_
_Updated title and description (miniche, 30/07/2013)_

PR #58, opened 2013-03-29
CentOS distributions Support

Guimove:


Guimove:
Ping @Ninir & @miniche

centralreport/cr/tools.py
```diff
@@ -230,3 +232,9 @@ def determine_current_host():
             elif os.path.isfile('/etc/redhat_version'):
                 # RedHat!
                 Config.HOST_CURRENT = Config.HOST_REDHAT
+            elif os.path.isfile('/etc/arch-release'):
+                # ArchLinux!
+                Config.HOST_CURRENT = Config.HOST_ARCH
+            elif os.path.isfile('/etc/redhat-release'):
```
miniche: To be sure this is a CentOS distribution, isn't it better to check the content of this file?

E.g: by adding a test on the return of this command : `cat /etc/redhat-release | awk '{print $1}'` (must return "CentOS") or anything else, if you have a better idea.

Guimove: After some research, why not simply use 

``` Python
import platform
platform.linux_distribution()
```

For all Linux OS ? 

It returns for a 

CentOS : 

``` Python
('CentOS', '6.4', 'Final')
```

Debian : 

``` Python
('debian', '6.0.7', '')
```

Work well, but we have to check first the version of Python because in python < 2.6 we have to use platform.dist
Something like : 

``` Python
try:
    import platform.linux_distribution as dist  # Python > 2.6
except: import platform.dist as dist  # Python < 2.6

dist()
```

With this, I think we can now get all linux distros.

miniche: Yeah, that could be the best solution!
I will try this on all my VMs this afternoon.

Guimove: I already tested in some VM in my work, all working well.

I already use something like that in collector.py 

``` Python
# Getting OS Name and OS version
        if sys.version_info[:2] < (2, 6):  # Python < 2.6
            from platform import dist
        else:
            from platform import linux_distribution as dist

            # Getting OS Name and OS version
        try:
            os_name = dist()[0]
            os_version = dist()[1]
        except:
            os_name = 'Linux'
            os_version = ''
```

miniche: Exact. I think it's a better solution to detect the OS instead of my first work :+1:

Guimove: Here is a list of returns I get during my tests : 

CentOS : 

``` Python
('CentOS', '6.4', 'Final')
```

Debian : 

``` Python
('debian', '6.0.7', '')
```

RedHat : 

``` Python
('redhat', '5.8', 'Tikanga')
```

miniche: For me,

Ubuntu returns:

``` bash
('Ubuntu', '11.10', 'onelric')
```

Fedora returns:

``` bash
('Fedora', '16', 'Verne')
```

Guimove:
Ok, I'll make the changes, according to this debate, tonight.

miniche:
:+1: 

Next step: Update bash functions and Python collectors to support these distros :relaxed:

Guimove:
Ok I'll work on that later this month

Guimove:
Ping @Ninir

miniche:
What you said? It will be ready for monday? Great! You're hired!
:D

Guimove:
Ping @miniche @Ninir

bash/utils.inc.sh
```diff
@@ -18,7 +18,7 @@ function getOS(){
 
     if [ "Darwin" == $(uname -s) ]; then
         CURRENT_OS=${OS_MAC}
-    elif [ -f "/etc/debian_version" ] || [ -f "/etc/lsb-release" ]; then
+    elif [ -f "/etc/debian_version" ] || [ -f "/etc/lsb-release" ] || [ -f /etc/redhat-release ]; then
         CURRENT_OS=${OS_DEBIAN}
```
miniche, line 40: Is it not be better to separate the Debian and the Redhat families by making a new OS_REDHAT const?
Commands are almost the same, but this can avoid your double detection in functions.inc.sh (l. 155 and l.314).

Guimove, line 40: Yep 
I'm already working on this.

This commit was temporary for testing purpose.

Guimove, line 40: I added : 

```
OS_CENTOS="CentOS"
OS_REDHAT="RedHat"
```

Arch will come after

Guimove:
For information, in this state, the web server will not work on CentOS because the routes library doesn't work on it.
No problem, this library is removed on PR #63

Ninir:
Seems good

Ninir:
@Guimove Please do some tests once again in order to be sure that everything is working fine since #63 has been merged.

Guimove:
Now I got this error : 

``` Bash

Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/centralreport/cr/threads.py", line 67, in run
    Checks.hostEntity = self.MyCollector.get_infos()
AttributeError: 'Checks' object has no attribute 'MyCollector'
```

I'll investigate with my Mac later, that's impossible to dev on my work computer running Windows :angry: 

For infomation : My centOS is running python 2.6.6 like my debian who is working

Ninir:
Having big screens but a shitty PC is not that good in fact... :smile: 

Isn't it possible to run a virtual machine?

miniche:
@Guimove, for your error, look at the constructor in the Config class (threads.py module).

Guimove:
Yup this is were I was looking for.

Guimove:
Ok for review @Ninir @miniche (Please, make it be the last :trollface:)

centralreport/cr/threads.py
```diff
@@ -43,7 +43,8 @@ def __init__(self):
 
         if Config.HOST_CURRENT == Config.HOST_MAC:
             self.MyCollector = collectors.MacCollector()
-        elif (Config.HOST_CURRENT == Config.HOST_DEBIAN) or (Config.HOST_CURRENT == Config.HOST_UBUNTU):
+        elif (Config.HOST_CURRENT == Config.HOST_DEBIAN) or (Config.HOST_CURRENT == Config.HOST_UBUNTU) or \
+                (Config.HOST_CURRENT == Config.HOST_CENTOS):
             self.MyCollector = collectors.DebianCollector()
```
miniche, line 48: If the collector is the same for Debian/Ubuntu/CentOS (and Redhat in this case?), wouldn't be better to rename the DebianCollector to LinuxCollector?
Just an idea :)

miniche:
Is it just working for CentOS or also for Redhat/Fedora?

Guimove:
We need to talk about that, I think we have to refactor the Config.HOST_CURRENT because we'll have conditions like that : 

``` Python
if (Config.HOST_CURRENT == Config.HOST_DEBIAN) or (Config.HOST_CURRENT == Config.HOST_UBUNTU) or  (Config.HOST_CURRENT == Config.HOST_CENTOS) or (Config.HOST_CURRENT == Config.HOST_FEDORA) or (Config.HOST_CURRENT == Config.HOST_REDHAT) or ...
```

I think we have to decompose like : 

``` Python
HOST_CURRENT = HOST_MAC
HOST_CURRENT = HOST_LINUX
```

And another variable like : 

``` Python
# Distro specific 
LINUX_TYPE = HOST_DEBIAN
LINUX_TYPE = HOST_UBUNTU
LINUX_TYPE = HOST_CENTOS
LINUX_TYPE = HOST_FEDORA

# or Distro family
LINUX_TYPE = DEBIAN_FAMILY  # Debian, Ubuntu...
LINUX_TYPE = REDHAT_FAMILY  # CentOS, Fedora...
```

We this we can do 

``` Python
if Config.HOST_CURRENT == Config.HOST_MAC:
              # Do something...
elif Config.HOST_CURRENT == Config.HOST_LINUX:
              # Do something...
```

With in specific situation : 

``` Python
if Config.HOST_CURRENT == Config.HOST_MAC:
              # Do something...
elif Config.HOST_CURRENT == Config.DEBIAN_FAMILY:
              # Do something...
elif Config.HOST_CURRENT == Config.REDHAT_FAMILY:
              # Do something...
```

miniche:
Yes, we must have more granularity. It's true, we can using families and types.
- CURRENT_OS can be:

```
OS_MAC = "Mac"
OS_DEBIAN = "Debian"
OS_UBUNTU = "Ubuntu"
OS_FEDORA = "Fedora"
OS_REDHAT = "Redhat"
OS_CENTOS = "CentOS"
OS_ARCH = "ArchLinux"
```
- CURRENT_FAMILY can be:

```
FAMILY_MAC = "Mac"
FAMILY_DEBIAN = "Debian"
FAMILY_REDHAT = "Redhat"
```
- CURRENT_TYPE can be:

```
TYPE_MAC = "Mac"
TYPE_LINUX = "Linux"
```

miniche:
Yeah, working well on my virtual machine! (CentOS 6.4) Great :)
capture decran 2013-05-03 a 00 02 47

Guimove:
Their's a risk of confusion, we have to distinct FAMILY vs OS
- CURRENT_FAMILY can be:

``` Bash
FAMILY_MAC = "Mac Family"
FAMILY_DEBIAN = "Debian Family"
FAMILY_REDHAT = "Redhat Family"
```

or something like that

miniche:
Yes!

centralreport/cr/threads.py
Ninir, line 48: It has to be renamed if it is used by Linux distros IMO

centralreport/cr/tools.py
```diff
@@ -219,17 +224,35 @@ def determine_current_host():
...
 
             # Utilisation de la liste de Novell pour reconnaitre des distrib Linux
-            # http://www.novell.com/coolsolutions/feature/11251.html
+            # http://www.novell.com/coolsolutions/feature/11251.html - Deprecated here
 
-            if os.path.isfile('/etc/lsb-release'):
-                # Ubuntu!
+            # Getting OS Name and OS version
+            if sys.version_info[:2] < (2, 6):  # Python < 2.6
+                from platform import dist
```
Ninir, line 231: Are we sure that this will work on Python 2.5 too? (introduced by #71)

Guimove, line 231: Yes ! 
The dist module is the old version of linux_distribution, dist was deprecated in python 2.6.

So here I check the python version to load the good "distribution" module.

Ninir, line 231: Nice :)

Guimove:
Can we merge that ?

Ninir:
You should merge develop in it in order to have Travis CI build running fine :)

Guimove:
Yep but I'll wait for @miniche to merge some PR

miniche:
All needed branches have been merged in "develop", but I think you will have a lot of conflicts.

In your work, you have updated some functions to detect CentOS, but also a Redhat distribution.
If you want to focus on CentOS and implementing Redhat/Fedora later, I think it's better to remove (or comment) the parts on Redhat detection. Otherwise, somebody can install CentralReport on a unsupported distribution.

Guimove:
I merged 'develop'.

I removed RedHat support.

Working for me

miniche:
I will review this pull request tomorrow ;-)

miniche:
I think your last merge (develop into this branch, b9bb2ab) erase a lot of code in these files:
- bash/functions.inc.sh
- centralreport.py

You should check this. I can be easier to locate your updates and use a clean copy of these files from the develop branch.

bash/utils.inc.sh
```diff
@@ -38,6 +38,15 @@ function getOS(){
         CURRENT_OS=${OS_MAC}
     elif [ -f "/etc/debian_version" ] || [ -f "/etc/lsb-release" ]; then
         CURRENT_OS=${OS_DEBIAN}
+    elif [ -f "/etc/redhat-release" ]; then
+        OS=`cat /etc/redhat-release | awk {'print $1'}`
+        if [ ${OS} != "CentOS" ]; then
+            CURRENT_OS=${OS_REDHAT}
```
miniche: If you keep this here, in addition with the condition `if [ ${CURRENT_OS} == "${OS_OTHER}" ];` in the install.sh script, CentralReport installer will accept to start the installation procedure on a RedHat distribution (and some unexpected behaviors).

Just an idea to improve this function:

``` bash
function getOS(){

    CURRENT_OS=${OS_OTHER}

    if [ $(uname -s) == "Darwin" ]; then
        CURRENT_OS=${OS_MAC}
    elif [ -f "/etc/debian_version" ] || [ -f "/etc/lsb-release" ]; then
        CURRENT_OS=${OS_DEBIAN}
    elif [ -f "/etc/redhat-release" ]; then
        OS=`cat /etc/redhat-release | awk {'print $1'}`
        if [ ${OS} == "CentOS" ]; then
            CURRENT_OS=${OS_CENTOS}
        fi
    fi
}
```

centralreport/cr/tools.py
```diff
@@ -219,17 +224,35 @@ def determine_current_host():
...
+                # Ubuntu
                 Config.HOST_CURRENT = Config.HOST_UBUNTU
-            elif os.path.isfile('/etc/debian_version'):
-                # Une Debian pure et dure dans ce cas!
+            elif dist()[0] == "debian":
+                # Debian
                 Config.HOST_CURRENT = Config.HOST_DEBIAN
-            elif os.path.isfile('/etc/fedora-release'):
-                # Fedora!
+            elif dist()[0] == "Fedora":
```
miniche: It's the same here as the bash script for OS detection. In this case, you will able to launch CR on a not compatible OS.
To avoid loosing your work, you can comment the Fedora, Redhat and ArchLinux detectors and just keep the CentOS one.

bash/utils.inc.sh
Guimove: Did you read this function globaly ? 
My function already make the difference between Centos and RedHat Like.

This function only set the CURRENT_OS var but only Debian, Ubuntu and CentOS will pass (cf the installer)

centralreport/cr/tools.py
Guimove: Like my previous comment, this function only set the OS, I think we can let this detection as we only let Debian, Ubuntu and CentOS install and use CR.

bash/utils.inc.sh
miniche: :)

In your 'installer.sh', you have updated the check on the current host to the following:

```
+if [ ${CURRENT_OS} == "${OS_OTHER}" ]; then
     printBox red "ERROR!| \
                   The install is only designed for Mac OS, Debian and Ubuntu.| \
                   Support for other OS will come soon!"
```

With your detection code, an Redhat-like distribution will have the variable CURRENT_OS defined to OS_REDHAT, and not to "OS_OTHER".
In this case, in the installer.sh, the condition CURRENT_OS==OS_OTHER is not filled, and the installer can continue.

Or, perhaps, I've missed something...

Guimove: Oops sorry I missed that.
Thanks (long time without working sorry)

centralreport/cr/tools.py
miniche: In centralreport.py, your test is:

```
if Config.HOST_CURRENT != Config.HOST_OTHER:

            log.log_info('%s detected. Starting ThreadChecks...' % Config.HOST_CURRENT)
            CentralReport.checks_thread = threads.Checks()  # Launching checks thread
        else:
            is_error = True
            log.log_critical('Sorry, but your OS is not supported yet...')
```

If you put `Config.HOST_CURRENT = Config.HOST_FEDORA` (idem for ArchLinux and Redhat), the HOST_CURRENT is different from HOST_OTHER, and CentralReport will accept to start itself. Or, it's the same, i've missed something in this case.

bash/utils.inc.sh
miniche: Errare humanum est :stuck_out_tongue_winking_eye:
And welcome back :p

Guimove:
Lets review, I think everything is good regarding your comments.

bash/functions.inc.sh
```diff
@@ -183,17 +183,24 @@ function copy_init_file(){
...
-        RETURN_CODE="$?"
+            RETURN_CODE="$?"
 
-        if [ ${RETURN_CODE} -ne "0" ]; then
-            logError "Error registering the startup script with update-rc.d (Error code: ${RETURN_CODE})"
-            return 2
+            if [ ${RETURN_CODE} -ne "0" ]; then
+                logError "Error registering the startup script (Error code: ${RETURN_CODE})"
+                return ${RETURN_CODE}
+            else
```
miniche: Useless `else` here. We already return the '0' value at the end of this function ;-)

bash/functions.inc.sh
```diff
@@ -183,17 +183,24 @@ function copy_init_file(){
 
         if [ ${RETURN_CODE} -ne "0" ]; then
             logError "Error copying the startup script at ${STARTUP_PLIST} (Error code: ${RETURN_CODE})"
-            return 1
-        fi
+            return ${RETURN_CODE}
```
miniche: It must be `return 1` (as described in the function docstring)

bash/functions.inc.sh
```diff
@@ -183,17 +183,24 @@ function copy_init_file(){
...
-        update-rc.d centralreport defaults
-        RETURN_CODE="$?"
+            RETURN_CODE="$?"
 
-        if [ ${RETURN_CODE} -ne "0" ]; then
-            logError "Error registering the startup script with update-rc.d (Error code: ${RETURN_CODE})"
-            return 2
+            if [ ${RETURN_CODE} -ne "0" ]; then
+                logError "Error registering the startup script (Error code: ${RETURN_CODE})"
+                return ${RETURN_CODE}
```
miniche: It must be `return 2` (as described in the function docstring)

miniche:
I seems to be good, but I will test it tomorrow for the final approuvement ;-)

Guimove:
Review ?   :turtle::dash:  :speak_no_evil:

centralreport/cr/collectors.py
```diff
@@ -425,7 +425,7 @@ def get_disks(self):
         list_disks = host.Disks()  # Return new entity
 
         for i in range(1, len(df_split)):
-            if df_split[i].startswith('/dev/'):
+            if df_split[i].startswith(('/dev/s', '/dev/h')):
```
miniche: I've made some tests about that. It seems that CentOS have the LVM functionality enabled by default (official doc: http://www.centos.org/docs/5/html/Cluster_Logical_Volume_Manager/)

In this case, the "df" command doesn't return the name of the physical partitions, but the name of logical partitions. Consider my virtual machine on which I've chosen all default values during the installation of CentOS.
I have one physical disk of 20 GB, according to this screenshot:
disk centos

The "df" command return this:
centos df

We can see a physical partition, the "/dev/sda1", with a size of 500 MB (approximation, 49588 KB exactly). It's only mounted as "/boot". Good.
However, as you can see, the "/" mount point which contains all the data is mapped with /dev/mapper/vg_centos-lv_root.
In reality, this **logical** partition is linked to `/dev/dm-0`.

With this test (`if df_split[i].startswith(('/dev/s', '/dev/h')):`), we can only see physical partition, and not logical ones.
And I think it's a little complicated on a system with the "lvm" tools enabled by default :P

NB: We are also working on the disks in #52, with the UUID detection. In this case, we should use the "name" as "/dev/mapper/vg_centos-lv_root", and "display_name" as "dm-0". Moreover, it can be useful to get the "mount point" to identify more quickly the disks/partitions.

Guimove:
This can be part of an other issue ?

miniche:
Nop :smile: 
We can't have partial or wrong data. If we omit the lvm partition, this collector returns a wrong list of disks. I think it's more important to have an raw name (like "mapper/vg_centos-lv_root") instead of nothing.
We can merge this PR before the #52, and I will update my data collector for Debian to find the UUID and the "display_name" of these partitions.

However you can open an issue to adding more data in the "disk" entity, like the "mounted point".

miniche:
And you can open an other issue for the Redhat/Fedora support, with the new OS detection (with the OS and the family, see: https://github.com/CentralReport/CentralReport/pull/58#issuecomment-17367588)

:smile:

Guimove:
Here is an issue explaining why I ignored logical partitions for the moment : #78

miniche:
Could you update the "NEWS" file with this CentOS support? :)

miniche:
Seems good for me. Tested on CentOS 6.4 :+1:

Guimove:
NEWS Updated, good for me too. 

Waiting for your f*****g approval. :tongue:

NEWS
```diff
@@ -8,6 +8,9 @@
 Version Alpha 0.3.0 (Not released yet)
     Currently in development
 
+    - Feature #58 : Can now detect if it's a CentOS
```
Ninir: "if it's a CentOS" -> if what is a CentOS? ;)
"We" know, external people may not. I would go for "CentOS distributions Support", thus renaming the PR. What do you think?

bash/vars.inc.sh
```diff
@@ -13,6 +13,9 @@
 CURRENT_OS=
 OS_MAC="MacOS"
 OS_DEBIAN="Debian"
+OS_CENTOS="CentOS"
+OS_REDHAT="RedHat"
+OS_OTHER="Not Supported OS"
```
Ninir: `Unsupported` seems more appropriate

centralreport/cr/tools.py
Ninir, line 231: Regarding Python 2.5 issues with JSON, etc... do we still want to support it?

miniche, line 231: Nope! (#76)

Guimove:
Merge ?

miniche:
:+1:

centralreport/cr/tools.py
```diff
@@ -219,17 +224,35 @@ def determine_current_host():
...
-                Config.HOST_CURRENT = Config.HOST_REDHAT
+            #elif dist()[0] == "Fedora":
+                # Fedora
+            #    Config.HOST_CURRENT = Config.HOST_FEDORA
+            #elif dist()[0] == "redhat" or dist()[0] == 'Red Hat Enterprise Linux Server':
+                # RedHat
+            #    Config.HOST_CURRENT = Config.HOST_REDHAT
+            #elif os.path.isfile('/etc/arch-release'):
+                # ArchLinux
+            #    Config.HOST_CURRENT = Config.HOST_ARCH
```
Ninir, line 252: As you are commenting these lines, the tests are uneeded. Python will evaluate them for nothing, thus losing time. Why not adding them only when needed?

centralreport/cr/tools.py
```diff
@@ -219,17 +224,35 @@ def determine_current_host():
...
+            elif dist()[0] == "CentOS":
+                # CentOS
+                Config.HOST_CURRENT = Config.HOST_CENTOS
+            else:
+                # No match, default
```
Ninir, line 257: This seems pretty uneeded :P

Ninir:
Seems fine :) :+1:

Guimove:
They are commented and not used for the moment because I don't want to lose this work.
This will be uncommented progressively, when a new OS is supported.

Ninir:
@Guimove having code that is evaluated for nothing is a bad idea as this is just slowing down this part of the Project. Btw, github Will still give you the history of the file. This is only an information, this is not really critical ;)
Good job!

PR #62, opened 2013-04-07
Fixed imports in the dev tools script

miniche:
The import of the "utils.inc.sh" was massing in the dev tools.
It works now, tested on OS X.


Guimove:
Ok tested on both deb and mac

miniche:
Just waiting for the "Ninir +1" :D

(PS: the order of imports is important, it can't by alphabetical order)

Ninir:
:+1:

PR #44, opened 2013-02-27
Improving CentralReport security and integration on operating systems

miniche:
This pull request will change many elements in CentralReport. Summary:
- [x] Adding a system user to secure CentralReport daemon.
- [x] Adding a launch script in /usr/local/bin, which will be included in the PATH.
- [x] All python modules will go to /usr/local/lib.

Now, with details:
## CentralReport user

For the moment, CentralReport is executed with root permissions. It isn't really secure, we can't say otherwise. So, to secure scripts, CentralReport will have his own user, with only few rights in the system.
- [x] Make some research
- [x] Add the new user on Debian
- [x] Add the new user on Mac OS
- [x] Update all install/uninstall functions
- [x] Update rights on /usr/local/lib/centralreport (see "Launch script")
- [x] Make a new folder in /var/log and update rights
- [x] Make a new folder in /var/run and update rights
- [x] Update documentation
## Launch script

For the moment, we must start or stop CentralReport with a long command: `sudo python /usr/local/bin/centralreport/centralreport.py start`. Yeah... Boring.
By adding a new script, it will be easier. The new command can look like: `centralreport start`. Better, no?
- [x] Creation of the new binary script (CentralReport user is required)
- [x] Update install and uninstall bash scripts for all OS (copy, delete and chmod +x)
- [x] Update init script (init.d for Debian and LaunchDaemon for Mac)
- [x] Update documentation
## Python modules

Put all python scripts and modules in /usr/local/lib is more logical according to me.
- [x] Update all bash functions (install/uninstall)
- [x] Update documentation

PR #63, opened 2013-04-07
Removed "Routes" library

miniche:
Routes requires "setuptools" installed on each host to work. After the security release (PR #44), this library can't work any more.
This is an "emergency" fix. A better solution will be proposed in a next release (custom route object? Use another http framework instead of CherryPy? …)

Tested on OS X and Debian


miniche:
A little precision:

I have created an **Api** class in the server.py module. It handles all calls in the /api route.
In this class, we have the `check` method to handles the "/api/check" route.

The `check` method have the "route" argument. It can be `date`, `full` or `disk` and delivers the right content to match with with all existing routes (/api/check/date, /api/check/full and /api/check/disk)

Guimove:
Everything is working well :+1:

miniche:
Ready to be merged, waiting for the @Ninir review :)

centralreport/web/server.py
```diff
@@ -116,13 +102,185 @@ def stop(self):
...
+
+    """
+        This class handles the "/api" route in the webserver
+        Gives raw data about the current host and the last check
+    """
+
+    @cherrypy.expose()
+    def check(self, route):
+        """
+            Gets informations about checks
```
Ninir: `Gets information` without a s

centralreport/web/server.py
```diff
@@ -116,13 +102,185 @@ def stop(self):
...
+                tmpl_vars['last_timestamp'] = '0'
+                tmpl_vars['last_fulldate'] = 'Never'
+            else:
+                tmpl_vars['last_timestamp'] = datetime_to_timestamp(Checks.last_check_date)
+                tmpl_vars['last_fulldate'] = Checks.last_check_date.strftime("%Y-%m-%d %H:%M:%S")
+                tmpl_vars['current_timestamp'] = datetime_to_timestamp(datetime.datetime.now())
+
+                # CPU Check informations
+                if Checks.last_check_cpu is None:
+                    tmpl_vars['cpu_check_enabled'] = 'False'
```
Ninir: What happens if you pass `False` rather than `'False'`?

centralreport/web/server.py
```diff
@@ -116,13 +102,185 @@ def stop(self):
...
+                        tmpl_vars['swap_percent'] = int(Checks.last_check_memory.swap_used) * 100 / int(
+                            Checks.last_check_memory.swap_size)
+                        tmpl_vars['swap_used'] = text.convert_byte(Checks.last_check_memory.swap_used)
+
+                        tmpl_vars['swap_free'] = text.convert_byte(Checks.last_check_memory.swap_free)
+                        tmpl_vars['swap_size'] = text.convert_byte(Checks.last_check_memory.swap_size)
+
+                        # On Mac, the swap is unlimited (only limited by the available hard drive size)
+                        if Checks.last_check_memory.swap_size == Checks.last_check_memory.total:
+
```
Ninir: Useless empty line

centralreport/web/server.py
```diff
@@ -116,13 +102,185 @@ def stop(self):
...
+                            tmpl_vars['swap_configuration'] = 'unlimited'
+                        else:
+                            tmpl_vars['swap_configuration'] = 'limited'
+
+                        if isinstance(tmpl_vars['swap_percent'], int):
+                            if int(tmpl_vars['swap_percent']) >= int(Config.get_config_value('Alerts', 'swap_alert')):
+                                tmpl_vars['swap_state'] = 'alert'
```
Ninir: Instead of passing such strings (which could be mistyped), I would pass CONSTANTS defined at the top of this file. For instance:

``` python
STATE_ALERT = 'alert'
STATE_OK = 'ok'
STATE_WARNING = 'warning'
```

and then reuse them where needed

Ninir:
We should definitely implement tests but for now, this seems good :)

miniche:
Yeah, we must implement them. Quickly.
FYI, this internal webservice will be written from scratch in a next version. The actual solution is not really clean.

I'm applying your suggestions.

Ninir:
Let's go to Travissss yayyyy :smile: :D

centralreport/web/server.py
miniche: It's the same. Do you think it's better to use `False`?

Ninir: Dunno :/ 
There are less bytes used, for sure :P

miniche: Good argument! :D
CherryPy + Jinja update all dictionary data in strings for the output, so it can be better to use strings directly. Hum... Yeah. Perhaps we can win 0,02 ms during the processing. Or not ^^

centralreport/web/server.py
miniche: Hum... The empty line have been removed in the commit 48937e9, but Github hadn't mark your comment as "outdated".

miniche:
Ok, tested on Debian and OS X tonight, it works.

PR #67, opened 2013-04-17
Installation without interaction and Travis CI integration

Guimove:
- [x] Add Travis status button on README.md
- [x] Configure Travis
- [x] Add an option to launch the install without interaction


Ninir:
:+1: Great idea :)

miniche:
Yeah, if we have valid Travis builds ^^

Ninir:
The Travis hook has been deactivated, so it's not an issue anymore!
I'll add the configuration for this one if we do need to use Travis :)

(Or you could even do it by yourselves if you want to ;) )

Guimove:
Finnaly Travis will not be used for this project because of the complex process.
Can we delete this branch ?

Ninir:
The complex process? which is...? testing on several distributions?

Guimove:
Installation is in bash, all process are in python, and we have to check on multiple OS
I don't know if everything is really possible with Travis

Ninir:
Setting the language to python will still allow you to have the bash stuff. You can do several actions before launching your script so basically, everything is possible.

Not using Travis IS a shame IMO as you can run at least tests for one distrib.

Let's try to use it.

Ninir:
And if you don't want to, I'll do it on my own... so whatever, you can merge or not, I don't really care in this case.

Guimove:
Ok I do not know.
I thought we were limited to a single language.

Let's try it !

You can forget my previous comments

Ninir:
One example could be to do the following:

``` yml
language: bash

before_install:
- sudo apt-get update
- sudo apt-get install pythonX
```

This way, you are controlling the way you want to try your project, and then you can run a test

Ninir:
Here it works :)

.travis.yml
```diff
@@ -0,0 +1,12 @@
+--- 
+after_failure: 
+  - "cat /tmp/centralreport_error.log"
+after_script: 
+  - "service centralreport status"
+branches: 
+  only: feature-travis-in-readme
+language: python
```
Ninir, line 8: IMO, you should set it as bash, and then run in `before_script`:

``` yml
before_script:
 - apt-get update
 - apt-get install pythonX
```

Guimove:
It's now working like I want, but I have to tune some settings 
2013-04-25 14-18-29_Travis CI - Free Hosted Continuous Integration Platform for the Open Source Comm

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+  - "cat /tmp/centralreport_install.log"
+before_script: "ls ."
+branches: 
+  only: feature-travis-in-readme
+install: ~
+language: python
+python: 
+  - "2.6"
+  - "2.7"
+script: "sudo bash .travis.sh"
```
Ninir: You don't need to run this script (and then to create it), as the purpose of this `.travis.yml` file is well... to prepare the VM to build your project. Moreover, your script is just not readable IMO (don't take it hardly, it is not a mean comment ^^)

I would go for the following:

``` yaml
before_script:
  - ./install.sh --no-interaction
script:
  - service centralreport status --no-interaction
after_script:
  - ./uninstall.sh --no-interaction
```

Thus, we will add webservices calls/tests or any test suite we would like to run in the `script` section.

Finally, I find it more readable/maintainable and easy to do what the overall build will do in one file rather than checking the script-ception.

install.sh
```diff
@@ -77,12 +77,15 @@ if [ "$?" -ne 0 ]; then
...
-    logConsole " "
-    read -p "You will install CentralReport. Are you sure you want to continue? (y/N) " RESP < /dev/tty
+if [ "install" == ${ACTUAL_MODE} ] || [ "autoinstall" == ${ACTUAL_MODE} ]; then
+    if [ "autoinstall" == ${ACTUAL_MODE} ]; then
+        checkYesNoAnswer "Yes"
+    else
+        logConsole " "
+        read -p "You will install CentralReport. Are you sure you want to continue? (y/N) " RESP < /dev/tty
+        checkYesNoAnswer ${RESP}
+    fi
```
Ninir, line 89: Sorry to announce you that... we have to find a better way to proceed ^^

README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/CentralReport/CentralReport.png)](https://travis-ci.org/CentralReport/CentralReport)
```
miniche: This status can be current branch-related?

If the build fail on a dev branch, the build can be good on master for example. In this case, all readme in all branches displayed the fail status?

.travis.yml
miniche: "--no-interaction" seems to be a good choose for this argument name:
- It's clear and simple
- It can be the same for all scripts

And this can be used in the future for automatic deployments (yes, in a far far time ^^)

install.sh
miniche, line 89: Yeah, without yoda condition :smile:

miniche:
I've added some comments.
And.. Great work, Guimove!

README.md
Ninir: Yup

install.sh
Ninir, line 89: Not only :P

README.md
miniche: http://about.travis-ci.org/docs/user/status-images/#Build-Status-For-Specific-Branches

Perhaps we can only activate this feature for the master and the develop branches, but not all "features-" branches.
It can avoid some false errors when we are working on a specific point.

install.sh
miniche, line 89: All right ^^

README.md
Guimove: IMO, we can leave travis enabled on all branches as this will speed up the process of validation of PR

miniche: All right! But this image don't control Travis builds, it's only displaying their status.
We must have the Travis status on each PR, Github displays it in the PR summary.

But, if you add this image as it in the "Readme" file, all branches will be have the same.
If you are working on a "feature-" branch, with a lot of work to do and the build failed, the "Readme" on the master branch will also show a "Failed build". But the code in master still safe and efficient.

Do you understand what I mean?

miniche: Today is an excellent example. I've pushed commits in the "feature-online-installer" branch. The Travis build has failed. So, in your branch (feature-travis), the readme shows a "Build failed" image.

If we keep this image as it, all readme in all branches will displays a "Build failed", even in Master or Develop.
But this doesn't mean the master branch is not safe...

Github have a good integration with Travis build. Every commit displays its build status, and every Pull Request shows information about the lasted build in the branch. Adding, or not, this image doesn't change the travis activation.
I think we must keep Travis integration (with the Github hooks), but not this image, it can display wrong status on safe branches.

Guimove: We can add the status image for the master branch only like : 

```
[![Build Status](https://travis-ci.org/CentralReport/CentralReport.png?branch=master)](https://travis-ci.org/CentralReport/CentralReport)
```

With this we only have the status of the master branch

.travis.yml
miniche: @Ninir: After thinking about that, I think it will better to use a dedicated script to launch all tests. We have to following elements to check:
- Bash functions
- Python modules and functions
- Installer / Uninstaller
-  When CR is installed on the virtual machine: testing if the web part is working well / testing if binary script works ("centralreport stop", "centralreport restart" etc.)

According to the Travis documentation, we can't use more than one script in the "script" part (in .travis.yml).

To being cleaner, we can create a "tests" directory in the bash part, and another in the python part.
Each script will test one script/module (http://docs.python.org/2/library/unittest.html). And we can have a "global" script which launches all "sub" scripts. Travis just launches the "global" bash and the global python scripts.

miniche: Mmh... There is a little problem for the "--no-interaction" usage.
Bash includes a native function (getopts) to get all arguments of the script. But this function is restricted: there is only support for "single letters" arguments (like `-a` or `-b value`).

When I try with `--no-interaction`, bash sees many arguments... ("--", "-n", "-o", "-i", etc.).

If we want to use "--no-interaction", we can use a regext condition (if "--no-interaction" is located in arguments), but we never can use the getopts function.
If we wants to use simple argument (-z value) in the future, we should use another argument for this functionality (like "-s" (= silent) for example).

Ninir: `-s` (for silent) is fine. We could also use `-q` for quiet

miniche: Great :)
I've made changes in #71.

Ninir: > @Ninir: After thinking about that, I think it will better to use a dedicated script to launch all tests. We have to following elements to check:
> 
> Bash functions
> Python modules and functions
> Installer / Uninstaller
> When CR is installed on the virtual machine: testing if the web part is working well / testing if binary script works ("centralreport stop", "centralreport restart" etc.)

This is part of the tests: one test should check CR for a given behavior, another test should check it for another situation, etc. Tests are here to check your application. Give tests what you want, they will avoid you wasted time.

> To being cleaner, we can create a "tests" directory in the bash part, and another in the python part.
> Each script will test one script/module (http://docs.python.org/2/library/unittest.html). And we can have a "global" script which launches all "sub" scripts. Travis just launches the "global" bash and the global python scripts.

Well, first, we have to see how tests are run for bash and python. For instance, PHP is using phpunit. This way, we only need to launch `phpunit` in travis as a command.
Why not doing it for bash, something in the idea of:

``` yml
before_script:
- install.sh

script:
- bashunit # :smile:
- python -m unittest

after_script:
- uninstall.sh
```

This way, you are trying the installation (which is a pre-requisite). Then you check all your bash code, then your python code, then the uninstall.

Don't see any issue at the moment with this reasoning.

miniche: Ok.

miniche: Updated in #71.
Bash and Python tests will be created in a dedicated branch.

miniche:
I want you for the Pull Request review :D

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+
+# General declarations
+
+language: python
+python:
+  - "2.5"
+  - "2.6"
+  - "2.7"
+
+# Test definitions
```
Ninir, line 14: We are definitely adding unneeded comments... ^^

.travis.yml
```diff
@@ -0,0 +1,21 @@
+# ------------------------------------------------------------
+# Travis configuration for CentralReport
+# https://github.com/CentralReport
```
Ninir: This link should target the project, not the organization only

NEWS
```diff
@@ -4,6 +4,20 @@
...
 
+--------------------------------------------------------------
+Version Alpha 0.3.0 (Not released yet)
+    Currently in development
+
+    - Feature #67: Travis CI and silent installer/uninstaller
+        * Travis performs tests to ensure the code is sane, everytime
+          (https://travis-ci.org/CentralReport/CentralReport)
+        * New "-s" argument for the installer and the uninstaller scripts
+          When it used, tasks are fully automated, no user interaction required.
```
Ninir: `When used, tasks are fully automated and no user interactions are required.`
In english, 0 is plural. Only 1 is singular

NEWS
```diff
@@ -4,6 +4,20 @@
 # ------------------------------------------------------------
 
 
+--------------------------------------------------------------
+Version Alpha 0.3.0 (Not released yet)
+    Currently in development
+
+    - Feature #67: Travis CI and silent installer/uninstaller
+        * Travis performs tests to ensure the code is sane, everytime
```
Ninir: `Travis CI performs a Continuous Integration to ensure your codebase is never broken.`

install.sh
```diff
@@ -76,14 +70,26 @@ if [ "$?" -ne 0 ]; then
...
-if [ "install" == ${ACTUAL_MODE} ]; then
-    logConsole " "
-    read -p "You will install CentralReport. Are you sure you want to continue? (y/N) " RESP < /dev/tty
+if [ "${ARG_WRONG}" == true ]; then
+    printBox red "ERROR! Unknown argument| \
+                  Use: install.sh [-s]"
+else
+    INSTALL_CONFIRMED=false
+
+    # The "-s" argument allows silent installation, without any user interaction
```
Ninir: You can remove the comma here

uninstall.sh
```diff
@@ -15,14 +15,17 @@ source bash/functions.inc.sh
 
 source bash/010_uninstaller.inc.sh
 
+get_arguments $*
+
 # We are ready to uninstall CentralReport. Log this and print the header.
 logFile "-------------- Starting CentralReport uninstaller  --------------"
 
+# Cleaning console and display the header lightbox
```
Ninir: Why adding such comments? Isn't the script that clear? ;)

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
 # On debian, the current user must have administrative privileges.
 if [ "${CURRENT_OS}" == "${OS_DEBIAN}" ]; then
     if [[ $EUID -ne 0 ]]; then
-        ROOT_ERROR="You must be root to install CentralReport!"
-        logFile ${ROOT_ERROR}
-        printBox red ${ROOT_ERROR}
+        logFile "You must be root to install CentralReport!"
+        printBox red "You must be root to install CentralReport!"
```
Ninir: You should define a var containing this sentence to re-use it twice instead of duplicating the content, this will avoid potential missed c/p issues

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
...
+        if [ ${CURRENT_OS} == ${OS_MAC} ]; then
+            logInfo "Processing... CentralReport will be removed from this Mac."
 
-        logConsole " "
-        printBox red " Error uninstalling CentralReport!| \
-                       CentralReport may still be installed on this host| \
-                       | \
-                       Some logs have been written in ${ERROR_FILE}"
+            # On Mac OS, the user must have access to administrative commands.
+            # Testing if the "sudo" session still alive...
```
Ninir: `Checking whether the "sudo" session is still alive`

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
...
 
-    # Process to CentralReport installation...
-    if [ ${bit_error} -eq 0 ]; then
-        uninstall_cr
-        if [ "$?" -ne 0 ]; then
-            bit_error=1
-        fi
-    fi
+    if [ "${UNINSTALL_CONFIRMED}" == true ]; then
+        # 0 = no error during uninstall
```
Ninir: This is an Unneeded comment isn't it? :smile:

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
...
+            # Testing if the "sudo" session still alive...
+            sudo -n echo "hey" > /dev/null 2>&1
+            if [ "$?" -ne 0 ]; then
 
-    else
-        logFile "CentralReport has been deleted from your host."
+                echo -e "\n\nPlease use your administrator password to uninstall CentralReport on this Mac."
+                sudo -v
+                if [ $? -ne 0 ]; then
+                    logError "Enable to use root privileges!"
```
Ninir: Typo: `Unable`

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
...
-                       | \
-                       Some logs have been written in ${ERROR_FILE}"
+            # On Mac OS, the user must have access to administrative commands.
+            # Testing if the "sudo" session still alive...
+            sudo -n echo "hey" > /dev/null 2>&1
+            if [ "$?" -ne 0 ]; then
 
-    else
-        logFile "CentralReport has been deleted from your host."
+                echo -e "\n\nPlease use your administrator password to uninstall CentralReport on this Mac."
```
Ninir: `to uninstall CentralReport from this Mac`

uninstall.sh
```diff
@@ -47,81 +50,97 @@ fi
...
+        # Process to CentralReport installation...
+        if [ ${bit_error} -eq 0 ]; then
+            uninstall_cr
+            if [ "$?" -ne 0 ]; then
+                bit_error=1
+            fi
+        fi
+
+        if [ ${bit_error} -eq 1 ]; then
+            # We display a generic message: previous logs already the specific error message.
```
Ninir: Unneeded comment (and you have missing words in it)

uninstall.sh
Ninir: IIRC, Bash has a 0 return code on success and something else for the rest, right? if so, then the comment may be useful, otherwise it should be removed IMO.

Ninir:
Guys, I suggest to add comments only when they are useful. They are some which do not bring any value to scripts. Here is a proposal:
- **Useful**: Security purposes (with links), Root password demands, "logging the issue", etc.
- **Useless**: "Adding some space", "Cleaning console and display the header lightbox"

What do you think?

.travis.yml
miniche: Right

uninstall.sh
miniche: It's true, and it was already updated (see old lines 50 to 52)... Maybe a wrong handling during the merge (feature-travis-improvement in this branch)

uninstall.sh
miniche: Because I love writing useless things :smile: 
Will be improved.

uninstall.sh
miniche: Bash returns the error code. If a function/script returns "0", we don't have any error.
Will be improved.

.travis.yml
miniche: Fixed in 8c775e3

.travis.yml
miniche, line 14: These comments ("General declarations" and "Test definitions") are here to separate blocs.
For me, they are importants for following reasons:
- They justify the order and this cutting
- They implement the rule for the future variables (where to write them)

Alright, this file is pretty small. This is most true when you have big configuration files, without section description.

miniche:
It's true, we must keep only useful comments. In most cases, we will agree on which is useful and which is useless.
But some cases are more subjective.

It think it's important to define what types of comments are important and which are not:
- Justify a line / a section (Why did I do it like that?)
- Describe a complicated part
- Cutting the code in big parts

For example:

``` bash
# We display a generic message: previous logs already contain messages about the error.
```

is important to me. It's justify why we are not displaying more information.

``` bash
# Cleaning console and display the header lightbox
```

is useless. Must be destroyed with a bazooka. Or a nuclear weapon. Or just with a notepad. As you want :D

Ninir:
Bazooka is fine :P 

Let's merge this one! Good job guys :+1:

PR #70, opened 2013-04-28
Refactored online installer/uninstaller

miniche:
Before the version 0.1.0, I've already created a "online installer/uninstaller".
With this feature, anyone can install CentralReport on his host with a simple command in a terminal, without downloading the project before.

In the idea, you open the terminal, and you copy/past the following command:

``` bash
curl http://scripts.centralreport.net/install | bash
```

(This URL doesn't exist yet, but it will be the "tools/online_installer.sh" script)

This script...
- Checks the current OS
- Checks if Python is available
- Downloads a "full package"
- Unarchives the full package
- Updates the chmod on install.sh script (chmod +x install.sh)
- Executes the install.sh script
- Deletes the unarchived package
- Deletes the package downloaded

The full package must be generated before with the "tools/packager.sh" script.
This packager generate an archive (.tar.gz) with the complete CentralReport script (as you download it manually from Github in the master branch).
It must be available from the centralreport server, and the URL fixed in online installer and uninstaller.

With this feature, installing or deleting CentralReport from any host connected to Internet will be easy.

(PS: the same solution exists also for the uninstaller)


miniche:
We need three URLs for this feature:
- A URL where the full package is available
- A URL for the online_installer.sh script
- A URL for the online_uninstaller.sh script

I suggest:
- http://scripts.centralreport.net/install
- http://scripts.centralreport.net/uninstall
- http://files.centralreport.net/package.tar.gz

But, perhaps have you better ideas? :smile:

tools/online_installer.sh
```diff
@@ -7,20 +7,20 @@
...
 ARCHIVE="package.tar.gz"
-DIR="CentralReportIndev"
+DIR="CentralReportPackage"
 
 CURRENT_OS=""
 OS_MAC="MacOS"
 OS_DEBIAN="Debian"
 
-echo -e "\n\nWelcome to CentralReport one-line installer!"
+echo -e "\n\nWelcome to CentralReport online installer!"
```
Ninir, line 25: Not sure about the online part here. When referring to online, I think about an online installer, which is not really the case here. This is a command which downloads a package from the internet, right?

tools/online_uninstaller.sh
```diff
@@ -7,19 +7,21 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin uninstall on current host.
+# This script will download latest CentralReport version, and uninstall CR from this host
```
Ninir: `and uninstall the previous CR version from this host`

Btw, you don't need the coma after version

tools/online_installer.sh
```diff
@@ -7,20 +7,20 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin installation.
+# This script will download latest CentralReport version, and begin the installation.
```
Ninir: You don't need the coma after version

tools/online_uninstaller.sh
```diff
@@ -40,7 +42,16 @@ else
     exit 1
 fi
 
-# Downloading full package...
+# Testing if Python is available on this host
```
Ninir: `Checking whether Python is available on this host`

tools/online_uninstaller.sh
```diff
@@ -40,7 +42,16 @@ else
...
+echo -e "\nTesting Python version..."
```
Ninir: According to your comment, this is not the appropriate sentence to print IMO.
What about `Checking Python existence...`?

tools/packager.sh
```diff
@@ -10,54 +10,70 @@
...
+cd CentralReportPackage
 find . -name "*.pyc" -exec sudo rm -rf {} \;
 cd ../
 
-sudo rm -R CentralReportIndev/.git
-sudo rm -R CentralReportIndev/.idea
-sudo rm -R CentralReportIndev/.DS_Store
+logConsole "Removing useless dot files..."
+sudo rm -R CentralReportPackage/.git 2>/dev/null
+sudo rm -R CentralReportPackage/.idea 2>/dev/null
```
Ninir: This should not even be there in fact. It should be removed from the Github repo

tools/online_uninstaller.sh
Ninir: If it's an uninstaller... why should you download the latest CentralReport version?
(Didn't check how it works:) Is it meant to be used in another script?

tools/online_installer.sh
miniche, line 25: In this case... What is your definition of a "online installer"?

Yes, this script downloads a package, but do a little more than this. I've detailled all steps in the PR description ("This script..." part)

tools/online_uninstaller.sh
miniche: Because you're doing more than a simple folder deletion (root access, unregistering init script, etc.), and all this steps are executed by bash functions. 

We can also make a "lite" package, with just bash scripts, without the python part (just the "bash" folder and the uninstall.sh script). That could be great, no?

tools/packager.sh
miniche: Yes, but the packager take the current folder on your "dev" computer, not from the repository. And in this case, we have this `.idea` folder.
Another solution can be delete all dot files (.*).

Ninir: There are 2 ways to proceed here:
- Making a whitelist, blacklisting what you don't want
- Making a blacklist (saying that every files/dir should not be included), whitelisting what you want to add.

So instead of taring the whole dir, I would go for a selection in files/dir. 

What do you think?

tools/online_installer.sh
Ninir, line 25: I would say "online" installer for something that is done on a web page, online.

THere might be other words to go with here... let's find them! :smile:

miniche, line 25: Mmh ok, I understand what you mean.
Before, I'd used the "one-line" composed word because with this functionality, you just write one line in you terminal, and all the (hard) work is done.

If you find a better wording, I take! :stuck_out_tongue_winking_eye:

Ninir, line 25: A perfect example could be: CloudShare or WindowsAzure, you can install/configure your  VM online. Is it more accurate? :/
Don't have better examples atm :/

tools/packager.sh
miniche: Good idea for a blacklist. It's more flexible than a whitelist in our case.

We need two blacklists:
- For the "installer" package (with all scripts included)
- For the "uninstaller" package (with just files required to perform the uninstallation)

I will try to implement this before the end of the week.

tools/online_installer.sh
miniche, line 25: I think it will be hard to find the _good_ word :smile:

Ninir, line 25: Let's keep it for now, you got me ^^ (don't have a better proposition ^^)
However, **please add a todo** as a comment so that we know we should find a better way (maybe when we'll have a native english speaker :P)

tools/packager.sh
Ninir: Thank you, you're the man :P

miniche: Or perhaps a whitelist... I'm thinking about that.
The whitelist can contain files or folders, for more flexibilities, and it may be easier to maintain:
- It's easier to detect missing files/folders with tests instead of unused files/folders
- With a blacklist, it will be hard to predict all unwanted files (.idea, .eclipse, .DS_store, etc.) for all developers

What do you think about that?

miniche:
We have now two packages: the "installer" and the "uninstaller".
Each package contains only useful files for their respective missions.

After a long meditation, I've chosen to create a whitelist and a blacklist for both.
The whitelist is read first. It contains a list of folders or files needed. Each item is written on a new line, and they will be copied on a temporary folder (in /tmp).
The blacklist is read after the whitelist. All unneeded items will be deleted from the temporary folder.

Why the both?
In fact, the whitelist is useful to describe all items needed for the packages. For example, we want to keep the folder "centralreport/" and its content, recursively. But, some folders, we can found ".pyc" file (files compiled by Python: http://docs.python.org/release/1.5.1p1/tut/node43.html). These files can't be included in the final packages.
The blacklist can contain files, folders and bash regex (starting by a star (*)). With the rule *.pyc, all files with the pyc extension will be deleted from the temporary folder, used for the package generation.

Finally, the installer package size is 971 Ko, and the uninstaller size is only 13 Ko.

tools/packager/uninstaller_blacklist.txt
```diff
@@ -0,0 +1,12 @@
...
+# https://github.com/CentralReport
+# ------------------------------------------------------------
+
+# This file contains all scripts/files/folders not used in the installer.
+# All items are deleted from the final package.
+
+# The blacklist is read after the whitelist.
+
+*.pyc
+*.DS_Store
```
Guimove, line 12: The .idea folder is not present in this blasklist ?

Guimove:
For the rest that's ok

tools/packager/uninstaller_blacklist.txt
miniche, line 12: No, because the .idea is not copied by the whitelist.
The blacklist is only useful to delete unneeded files/folders from the temporary folder, generated reading the whitelist.

tools/packager/installer_blacklist.txt
```diff
@@ -0,0 +1,12 @@
+# ------------------------------------------------------------
+# CentralReport Unix/Linux installer whitelist
```
miniche: To me: It's the blacklist, not the whitelist!

tools/packager/uninstaller_blacklist.txt
```diff
@@ -0,0 +1,12 @@
+# ------------------------------------------------------------
+# CentralReport Unix/Linux installer whitelist
```
miniche: Wrong header: it's the uninstaller blacklist.

tools/packager/uninstaller_whitelist.txt
```diff
@@ -0,0 +1,17 @@
+# ------------------------------------------------------------
+# CentralReport Unix/Linux installer whitelist
```
miniche: Wrong header: it's the uninstaller whitelist.

Ninir:
@miniche Commenting your own Pull Request make us think you are schizophrenic... ^^

miniche:
It's so hard to be agree with oneself... :D
I've noticed this just before a meeting, it's just to avoid forgetting :)

Guimove:
This man is always on a meeting... :trollface:

miniche:
@Ninir, what do you think about the lists? And you can do a review if you want :P

tools/online_installer.sh
Ninir, line 25: ping @miniche

miniche, line 25: Oh... I seems that I forgot to do that...

miniche:
I have made some improvements for this feature.
Now, the "packager" have its own functions: it's cleaner, and we can catch more exceptions and errors.
The "online" scripts are now copied in the final folder (in /tmp/cr-package for now) to put them on the CRO server.

You can test this executing the following command:

```
curl charles-emmanuel.me/cr/install | bash
```

(and the uninstaller is available with `curl charles-emmanuel.me/cr/uninstall | bash`)

I'm waiting for your reviews and suggestions :)

miniche:
Hey guys, don't forget to review this pull request :)

Guimove:
In Debian, curl was not installed, I had to install it manually.

In all my VMs, I get this issue : 

```
curl: (6) Couldn't resolve host 'charles-emmanuel.me'
```

But a ping works :angry:

miniche:
Right!

This command may work for Debian/Ubuntu:

```
wget -qO- http://charles-emmanuel.me/cr/install | bash
```

I forgot to add this command in my previous comment...
Can you confirm this for me?

Guimove:
Nop not working too

capture_d_cran_27_05_13_23_50

miniche:
I see that you're using zsh, I have only tried with bash. I will do more tests on my VM with zsh in this case.

Guimove:
Are you sure about the URL ?

capture_d_cran_27_05_13_23_51

miniche:
My bad... I modified the Apache configuration on my server during the last week... Shame on me.
I will fix it!

Guimove:
:scream: :rage: 

I'll check tomorrow now.

We have to had the support for CentOS too now following thie PR #58

miniche:
:worried:

Fixed!
capture decran 2013-05-27 a 23 58 46

miniche:
> I'll check tomorrow now.
> We have to had the support for CentOS too now following thie PR #58

:+1: 
We can finish the PR #58, merging it in develop and update this pull request after to add the compatibility with CentOS. Do you agree?

Guimove:
Ok for that but after the merge of the CentOS branch, we'll have to integrate this OS in all futur devs.

Guimove:
Working fine for me now

miniche:
Cool :)
And what do you think about this feature? These main advantages, in my opinion: it's cross-platforms and it's very simple. No package to be downloaded manually, no complex operations...

The ultimate work should be integrate CR in the official linux repositories (Debian, Ubuntu...). We can install CR with a simple "apt-get install centralreport" :D

Ninir:
Please make the package :)

Guimove:
We can make us own debain repository...let's check how to do this.

Here is some exemple : 
Package : http://alp.developpez.com/tutoriels/debian/creer-paquet/
Repo : http://blog.jonliv.es/2011/04/26/creating-your-own-signed-apt-repository-and-debian-packages/

miniche:
@Ninir: I've made the package on my own server for testing purposes.
If we want to generate it on our server, we must be agree about the URLs (Please see: https://github.com/CentralReport/CentralReport/pull/70#issuecomment-17134733)

@Guimove: Yeah, but in this case, we must update our own repositories list in the configuration file on Debian, doesn't it? I think it's better to find a solution to integrate the CR debian package in the official repositories.

Ninir:
> We need three URLs for this feature:
> 
> A URL where the full package is available
> A URL for the online_installer.sh script
> A URL for the online_uninstaller.sh script
> I suggest:
> 
> http://scripts.centralreport.net/install
> http://scripts.centralreport.net/uninstall
> http://files.centralreport.net/package.tar.gz
> But, perhaps have you better ideas? 

Seems good :)

miniche:
This PR will be updated with the work on #58 (CentOS distribution support) in the next days.

miniche:
I've generated the packages and the scripts from this current branch.

You can now make tests on CentOS and OS X:

```
curl http://scripts.centralreport.net/install | bash
```

On Debian/Ubuntu (they doesn't have "curl" installed by default):

```
wget -qO- http://scripts.centralreport.net/install | bash
```

To uninstall CentralReport, just replace "install" by..."uninstall" in the URLs.

Please give me your feelings :)

Guimove:
Ok for me in all my VMs

miniche:
Great :)
@Guimove, do you agree with the proposed URLs?

Ninir:
@miniche "do" you agree*

Gauthier Wallet
Envoyé depuis mon iPhone.

miniche:
This is what I wrote... What is the problem? :D
(Thanks, it was a ugly mistake...)

Guimove:
IMO their is to much subdomains, why not simply using one only entry point for downloads ?
Like : 
''''
Http://dl.central...
''''

miniche:
Yes, it may be better to keep one subdomain. Why not for "dl.centralreport.net".

Ninir:
What about "static.centralreport.net"? many sites are using this approach for their static files (images, scripts, ...)

miniche:
In fact, I doesn't have a better idea. We can have a dedicated subdomain for the CentralReport packages (installer/uninstaller, Unix/Linux and Windows installer after), or having a more generic subdomain.

Wouldn't it be better to have a subdomain for this functionality to keep the files organized and a simple URL, easy to remember?

Just some ideas:
- curl dl.centralreport.net/install
- curl linux.centralreport.net/install
- curl bash.centralreport.net/install
- curl NoMoreIdea.centralreport.net/install

miniche:
Ping @Guimove @Ninir about https://github.com/CentralReport/CentralReport/pull/70#issuecomment-19427440

Guimove:
I prefere the "static" or "dl" only because we don't have to create and manage a lot of subdomains.
Let's vote ?

NEWS
```diff
@@ -19,6 +19,12 @@ Version Alpha 0.3.0 (Not released yet)
           All confirmations are validated by default.
           Useful to perform tests or for deployments.
 
+    - Feature #70: New online installer and online uninstaller
```
Ninir: `New online installer and uninstaller` is enough :)

NEWS
```diff
@@ -19,6 +19,12 @@ Version Alpha 0.3.0 (Not released yet)
...
+        * Install and uninstall CentralReport with a single line in a terminal.
```
Ninir, line 23: Install and/or uninstall

NEWS
```diff
@@ -19,6 +19,12 @@ Version Alpha 0.3.0 (Not released yet)
...
+          One remote script is available for the installer and another for the
```
Ninir: and another one

NEWS
```diff
@@ -19,6 +19,12 @@ Version Alpha 0.3.0 (Not released yet)
...
+          uninstaller. They download the corresponding package automatically
+          and they start the processes.
```
Ninir: and then start processes. (we don't need the "the" word as you don't provide the process list)

tools/online_uninstaller.sh
```diff
@@ -7,18 +7,22 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin uninstall on current host.
+# This script will download latest CentralReport version and uninstall CR from this host
```
Ninir: Does it really download the latest CentralReport version or does it download only the uninstaller? (Looking ath the URL_CR, it downloads only the latest uninstaller).

What about `This script will download the latest CentralReport uninstaller and process the uninstall`?

tools/online_installer.sh
```diff
@@ -7,20 +7,22 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin installation.
+# This script will download latest CentralReport version and begin the installation.
```
Ninir: See https://github.com/CentralReport/CentralReport/pull/70/files#r4719125 for the same idea

tools/online_uninstaller.sh
miniche: It's true :)

tools/online_installer.sh
miniche: In this case, this script downloads the "installer package", which contains the lastest CentralReport version.

NEWS
miniche, line 23: Why `and/or`? I'm not sure that is a better formulation in this case (in English as well in French).

"You can install and uninstall CR with a sigle line in a terminal", in other words "You can do the both with your terminal". If you want to insist on the two separate scripts, I think it's better to use "or". 

(And for the story, the "et/ou" in French is not recommended, except in the familiar language and in newspapers: http://www.druide.com/enquetes/la-pseudo-conjonction-etou
In more formal writing, it must be avoided.

In English, it seems to be the same:
http://billbennett.co.nz/2009/09/26/writing-avoid-andor-slashes/
http://grammar.ccc.commnet.edu/grammar/plague.htm
http://grammar.quickanddirtytips.com/and-or-singular-plural.aspx)

miniche:
After our discussion this afternoon about that, the three URL will be:
- static.centralreport.net/installer
- static.centralreport.net/uninstaller
- static.centralreport.net/cr_installer.tar.gz
- static.centralreport.net/cr_uninstaller.tar.gz

Do you agree?

Guimove:
Go for that

miniche:
Scripts and packages have been generated and they are available here: http://static.centralreport.net/.
I will update the documentation tonight.

(Packages have been created with the current version in the master branch, the 0.2.0)

miniche:
Documentation updated: http://docs.centralreport.net/unix-agent/install

Ready to be merged.

Guimove:
Ok for me

Ninir:
:+1:

miniche:
I have just a little problem with our online server tonight. With the command `curl http://static.centralreport.net/installer`, our web server returns "HTTP/1.1 406 Not Acceptable".
(But it works fine with wget...)

I'm trying to fix it.

Ninir:
Try using `-O` as in `curl -O http://static.centralreport.net/installer`. It will then download the file locally using the remote name.

Guimove:
or like in the doc : 

```
curl http://static.centralreport.net/installer | bash
```

miniche:
@Ninir It doesn't work for me:
capture

@Guimove I have the same error in this case:
capture2

And when we had tried this functionality (https://github.com/CentralReport/CentralReport/pull/70#issuecomment-19297597) with other URL, everything worked well...

Guimove:
After some search, it looks like apache (or the web serveur used by CRO) is blocking cURL form accessing static file.

You can bypass this security faking your user agent (crappy solution not usable) : 

```
curl http://static.centralreport.net/installer -A "Mozilla"| bash
```

or we should look with us hosting or mod.security configuration

miniche:
It seems that we have a problem with the actual configuration of our web server.
We are investigating about this.

Guimove:
IMO you can merge this branch as the problem is not in the code but it's only a server configuration mistak

Ninir:
Merging a branch means that the work behind will be available to everyone once it's in master :/

The issue spotted is about curl itself: last checks proved that wget was also issuing changing the user-agent to be the same used by curl.

We should really find the issue and fix it as it is blocking potentiel users IMO.

What do you think guys?

miniche:
I think we must wait to have a fully functional solution before merging this branch into develop.

IMO, our priorities are:
- Finish the work on the branch feature-webservices (PR https://github.com/CentralReport/CentralReport/pull/52)
- Finish the work on the branch feature-manager (PR https://github.com/CentralReport/CentralReport/pull/82)
- Solve this problem
- Release the version 0.3.0

We must speak about this issue and our "online architecture" together :wink:

miniche:
Guys,

Packages are now available at `http://static.centralreport.net`.
@Guimove, can you try on your VMs with **curl** following the documentation? (http://docs.centralreport.net/unix-agent/install)

Packages will be updated when the version 0.3.0 will be out.

Guimove:
That's running well in my VMs. :+1:

miniche:
Great :)
@Ninir, you must say "Ok". Just "Ok" :D

NEWS
```diff
@@ -19,6 +19,12 @@ Version Alpha 0.3.0 (Not released yet)
           All confirmations are validated by default.
           Useful to perform tests or for deployments.
 
+    - Feature #70: New online installer and uninstaller
+        * Install and uninstall CentralReport with a single line in a terminal.
+          One remote script is available for the installer and another one for
+          the uninstaller. They download the corresponding package automatically
+          and they start processes.
```
Ninir: `they start processing`. Btw, what about "They are in charge to automatically download the related package and they are handling the magic for you." :D

tools/online_installer.sh
```diff
@@ -7,20 +7,22 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin installation.
+# This script will download the latest CentralReport installer and process the install
```
Ninir: `and processes`

tools/online_uninstaller.sh
```diff
@@ -7,18 +7,22 @@
 # https://github.com/CentralReport
 # ------------------------------------------------------------
 
-# This script will download latest CentralReport version, and begin uninstall on current host.
+# This script will download the latest CentralReport uninstaller and process the uninstall
```
Ninir: `processes`*

tools/packager/functions.inc.sh
```diff
@@ -0,0 +1,217 @@
...
+                    logError "Error copying ${line}!"
+                    return 1
+                fi
+
+            else
+                logError "Missing file or directory: ${line}"
+                return 2
+            fi
+        fi
+    done
```
Ninir, line 47: To remove indentation to this part (making it more readable), I would suggest to do this:

``` bash
for line in $(cat "$1");
do
    if [[ ${line} == "#"* ]] || [[ ${line} == ";"* ]] ; then
        continue
    fi

    if [ -d ${line} ]; then
        sudo cp -R ${line} "${2}${line}"
        RETURN_CODE="$?"

        if [ "$?" -ne "0" ]; then
            logError "Error copying ${line}!"
            return 1
        fi

    elif [ -f ${line} ]; then
        sudo cp ${line} "${2}${line}"
        RETURN_CODE="$?"

        if [ "$?" -ne "0" ]; then
            logError "Error copying ${line}!"
            return 1
        fi

    else
        logError "Missing file or directory: ${line}"
        return 2
    fi
done
```

Do you get the idea? What do you think?

tools/packager/functions.inc.sh
```diff
@@ -0,0 +1,217 @@
...
+#
+function read_whitelist(){
+
+    for line in $(cat "$1");
+    do
+        if [[ ${line} != "#"* ]] && [[ ${line} != ";"* ]] ; then
+            if [ -d ${line} ]; then
+                sudo cp -R ${line} "${2}${line}"
+                RETURN_CODE="$?"
+                if [ "$?" -ne "0" ]; then
```
Ninir: Instead of `"$?"`, why not checking `$RETURN_CODE`?

(ima guessing it contains $?)

tools/online_installer.sh
miniche: I must be more careful about your proposals :P (https://github.com/CentralReport/CentralReport/pull/70#discussion_r4719125)

tools/packager/functions.inc.sh
miniche, line 47: Yes, it's clearer!

tools/packager/functions.inc.sh
miniche: Why? You ask me why? No explanation. Bad developers, perhaps :D

NEWS
miniche: Not serious enough. I love :D

tools/packager/functions.inc.sh
miniche, line 47: Modified in c697fe0.

miniche:
Anything else? :)

Ninir:
Seems ok :)

miniche:
Packages are already available at http://static.centralreport.net. They are based on the version 0.2.0.
They will be updated when we will release the version 0.3.0.

PR #71, opened 2013-05-01
[Suggestion] Travis implementation improvements

miniche:
Some improvements:
- Added one function to detect and parse all script arguments (like "-s", etc.)
- The silent installation/uninstallation are available with the "-s" argument
- Testing CentralReport on Python 2.5 with Travis

This pull request will be merged in #67


miniche:
Ready for the review :)

Guimove:
Seems good to me.
Well done @miniche 
Sorry for being late on the project, no internet for some days.

miniche:
Don't worry :)

Guimove:
Actually, the "service" command doesn't return an error code. 
I think, your should add a simple unit test with : 

In .travis.yml

``` Bash
before_install:
  - wget --no-check-certificate https://shunit2.googlecode.com/files/shunit2-2.1.6.tgz
  - tar xvzf shunit2-2.1.6.tgz shunit2
script:
  - ./.unittest.sh
```

And creat a unittest file

.unittest.sh

``` Bash
#!/bin/bash

test_service(){

  out=`service centralreport status`;

  assertEquals "code retour" "$?" "0"
  assertNotEquals "stdout" "CentralReport is not running" "${out}"
}

. ./shunit2/src/shunit2
```

It returns, when it works (with error code 0) : 

``` Bash
test_service

Ran 1 test.

OK
```

What do you think about that for the moment ?

We'll make advanced tests in a futur PR.

miniche:
We must define a good architecture for implementing all tests.
In my mind, we have three test types:
- Bash unittests (functions behavior)
- Python unittests (functions/module behavior)
- General behavior (CR installer/uninstaller, CR "start"/"stop"/"restart", CR webserver availability when it's installed), can be written in bash or python

We can write tests in different folders to keep all tidy:
- Bash unittests in ~/bash/tests/
- Python unittests in ~/centralreport/tests/
- General behaviors in ~/tests/

Putting the Python tests in the python package is mandatory by the unittest library, and they must be launched by a script at the top of the package (in ~/centralreport).

Guimove:
I was only talking about an "alpha" test to close this PR with a little of consistency because for the moment the travis build will success even if the CR service doesn't start.

This architecture of test will be discussed in an other PR, what do you think about that ?

Moreover, I think we should have only one "tests" folder with subfolders for bash and python, we should centralize the unittests. 
The tests are not useful for production, I think we shouldn't scatter them too mush in our package.

miniche:
Oh ok, I didn't understand. 

Ok to discuss about it in the future pull request. Just: for python tests, I've already tried to write them in the root folder, but it seems not working. After reading the unittest module documentation, it seems you must write test modules in the package. But perhaps I had missed something :P

For me, it's better to start writting tests in the future PR, not here. It isn't useful to write a (dirty) temporary test because we will start to write clean tests just after.
It permits to avoid messages like "Build failed" from Travis after every commit in every branch. Just my opinion.

miniche:
So, can I merge this into the Guimove's branch (feature-travis-in-readme)?
We can make the (ultimate) review in the other branch before merging in develop.

Waiting for your approval or comments :)

.travis.yml
```diff
@@ -2,12 +2,11 @@
...
+after_script: "sudo ./uninstall.sh -s"
+before_script: "sudo ./install.sh -s"
 language: python
-python: 
+python:
+  - "2.5"
   - "2.6"
   - "2.7"
-script: "sudo bash .travis.sh"
+script: "service centralreport status"
```
Ninir, line 17: Having the python versions between before_script and script is alphabetically good, but it has no sense in fact. It presumes you know where to find the keys ;)

Why not doing the logical order? Something in the idea of:

before
during
after

Do you feel better with the alphabetical or the logical order?

.travis.yml
```diff
@@ -2,12 +2,11 @@
...
-before_script: "ls ."
-branches: 
-  only: feature-travis-in-readme
-install: ~
+after_script: "sudo ./uninstall.sh -s"
+before_script: "sudo ./install.sh -s"
 language: python
-python: 
+python:
+  - "2.5"
```
Ninir, line 10: (Excellent idea ;))

install.sh
```diff
@@ -76,17 +70,27 @@ if [ "$?" -ne 0 ]; then
...
 
-# Check the actual mode.
-if [ "install" == ${ACTUAL_MODE} ] || [ "autoinstall" == ${ACTUAL_MODE} ]; then
-    if [ "autoinstall" == ${ACTUAL_MODE} ]; then
-        checkYesNoAnswer "Yes"
+if [ "${ARG_WRONG}" == true ]; then
+    printBox red "ERROR! Unknown argument| \
+                  Use: install.sh [-s]"
+else
+    # Must be "true" to perform the installation.
```
Ninir: Uneeded comment, the var name and value are obvious, aren't they? :smile:

uninstall.sh
```diff
@@ -54,74 +56,88 @@ if [ "${CURRENT_OS}" == "${OS_DEBIAN}" ]; then
...
+
+if [ "${ARG_WRONG}" == true ]; then
+    printBox red "ERROR! Unknown argument| \
+                  Use: uninstall.sh [-s]"
 else
-    logConsole " "
-    read -p "You will uninstall CentralReport. Are you sure you want to continue (y/N)? " RESP < /dev/tty
-    checkYesNoAnswer ${RESP}
-fi
+    # Must be "true" to perform the installation.
```
Ninir: to perform the installation? Btw, isn't that really obvious? :P

miniche:
```
Good to merge — The Travis build passed
This pull request can be automatically merged.
```

Travis + Pull Requests = :heart:

Issue #65, opened 2013-04-10
CR not handling closed ports - Debian


Azaret:
Hi,

When some ports are not open, Central Report run without warning the user. Also the exception fired by python when a port is closed seems to be send only on debug logs, and not on error logs.

```
[04/10/2013 09:42:41] INFO:      ------------------------------------------------
[04/10/2013 09:42:41] INFO:      CentralReport is starting...
[04/10/2013 09:42:41] INFO:      Current user: azaret
[04/10/2013 09:42:41] DEBUG:     Linux config
[04/10/2013 09:42:41] DEBUG:     Configuration file: Found. Reading it.
[04/10/2013 09:42:41] DEBUG:     Reading the config file...
[04/10/2013 09:42:41] INFO:      Debian detected. Starting ThreadChecks...
[04/10/2013 09:42:41] DEBUG:     ThreadChecks is starting...
[04/10/2013 09:42:41] DEBUG:     Interval between two checks: 60 seconds
[04/10/2013 09:42:42] INFO:      Enabling the webserver...
[04/10/2013 09:42:42] INFO:      CentralReport started!
[04/10/2013 09:42:42] DEBUG:     ---- New check -----
[04/10/2013 09:42:42] DEBUG:     Doing a CPU check...
[04/10/2013 09:42:43] DEBUG:     Doing a memory check...
[04/10/2013 09:42:43] DEBUG:     Doing a load average check...
[04/10/2013 09:42:43] DEBUG:     Doing a disk check...
[04/10/2013 09:42:43] DEBUG:     All checks are done
[04/10/2013 09:42:43] DEBUG:     Next checks in 60 seconds...
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/usr/local/lib/centralreport/web/server.py", line 94, in run
    cherrypy.server.start()
  File "/usr/local/lib/centralreport/libs/cherrypy-3.2.2.zip/cherrypy/_cpserver.py", line 151, in start
    ServerAdapter.start(self)
  File "/usr/local/lib/centralreport/libs/cherrypy-3.2.2.zip/cherrypy/process/servers.py", line 167, in start
    wait_for_free_port(*self.bind_addr)
  File "/usr/local/lib/centralreport/libs/cherrypy-3.2.2.zip/cherrypy/process/servers.py", line 410, in wait_for_free_port
[...]

PR #72, opened 2013-05-10
Added a check for the webserver port availability

miniche:
A function in the new "web" module checks if a port is used (on the local or a remote computer).

The test have been added when CentralReport is starting. In this case, the webserver is not loaded and an error is written in logs.
This solves the issue #65.


miniche:
Ready to be merged

NEWS
```diff
@@ -17,6 +17,10 @@ Version Alpha 0.3.0 (Not released yet)
           All confirmations are validated by default.
           Useful to perform tests or for deployments.
 
+    - Improvement #72: Added a check for the webserver port availability
+        * When the choosen port is already used on the current computer,
```
Ninir: Typo: `Chosen`

`When the chosen port is already in use on the current host, the webserver isn't started and an error is logged.`

centralreport/centralreport.py
```diff
@@ -93,12 +94,18 @@ def run(self):
...
-        # Is webserver enabled?
+        # Launching the internal webserver...
         if not is_error and text.convert_text_to_bool(Config.get_config_value('Webserver', 'enable')):
-            from web.server import WebServer
+            local_web_port = int(Config.get_config_value('Webserver', 'port'))
 
-            log.log_info('Enabling the webserver...')
-            CentralReport.webserver_thread = WebServer()
+            if not utils_web.check_port('127.0.0.1', local_web_port):
+                log.log_info('Enabling the webserver...')
```
Ninir: Are we really `enabling` the web server here or are we starting it?

centralreport/centralreport.py
```diff
@@ -93,12 +94,18 @@ def run(self):
...
+
+                # Importing the module here improve the memory usage
+                from web.server import WebServer
+                CentralReport.webserver_thread = WebServer()
+            else:
+                log.log_error('Error launching the webserver: port %s is already used on this host!' % local_web_port)
```
Ninir: I would write `in use` instead of used

centralreport/cr/utils/web.py
```diff
@@ -0,0 +1,33 @@
+# -*- coding: utf-8 -*-
+
+"""
+    CentralReport - Web module
+        Contains useful functions around the web
```
Ninir: AFAIR, it is `from around`

centralreport/cr/utils/web.py
```diff
@@ -0,0 +1,33 @@
...
+
+    https://github.com/CentralReport
+"""
+
+import socket
+
+
+def check_port(ip, port):
+    """
+        Checks if the port is open on a specific IP
```
Ninir, line 15: `is opened`

Ninir:
Seems good

centralreport/centralreport.py
miniche: In this case, the verb "to enable" is a synonym of "to activate", isn't it?
We are not just starting the web server: we are loading the python module and all the libraries required, initialize CherryPy and after, we are starting the web server.

I think it's another a detail. What is the best word, in your opinion?

Ninir: Initialize fits better our needs then IMO

centralreport/cr/utils/web.py
miniche: True, thanks!

centralreport/cr/utils/web.py
miniche, line 15: Are you sure?
http://www.usingenglish.com/forum/ask-teacher/15771-open-opened-welcome-welcomed.html#post73496

Guimove:
:+1:

miniche:
Tested with Debian 6 and Ubuntu 12.04. Will be merged in the next seconds :)

PR #79, opened 2013-06-08
Removed support of Python 2.5

miniche:
All supported distributions (Ubuntu, Debian and CentOS soon) are shipped with Python 2.6 or Python 2.7.
Now, many libraries doesn't support Python 2.5 (such as "Routes" in our case), and we are using some modules (like the "json" one) only available since the version 2.6.

This pull request removes the support of Python 2.5 and adds two new checks: one in the installer and one in centralreport.py, before starting the daemon.

Please notice that CR is not fully compatible with Python 3. Many changes are needed, and they will be done in a next release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@
 
 language: python
 python:
-  - "2.5"
   - "2.6"
   - "2.7"
 
--- a/NEWS
+++ b/NEWS
@@ -27,6 +27,8 @@ Version Alpha 0.3.0 (Not released yet)
 
     - Fix #73: Removed old config.py script
 
+    - Fix #79: Removed support for Python 2.5 and older
+
 
 --------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


miniche:
Waiting for the review and your suggestions ;-)

Ninir:
Seems ok :)

Guimove:
I updated collector, ok for me

Ninir:
Go go go :)

miniche:
I just want to test on an old Linux distribution published with Python 2.4 or 2.5 before merging.
I will make the tests tomorrow ;)

miniche:
Seems good, tested on Debian 5.0 which was published with Python 2.5:
capture decran 2013-06-11 a 23 26 32

Works well with Python 2.6 and Python 2.7 (tested on Debian 6, 7 and OS X 10.8).

PR #80, opened 2013-06-09
[Test] Tornado web server

miniche:
Tornado is another lightweight Python web server (http://www.tornadoweb.org/).
Currently, we are using CherryPy (http://www.cherrypy.org/).

CherryPy seems to be a very good framework and Tornado too.
This branch is just here to discover the Tornado functionalities and to choose the best framework for us.


Ninir:
Why opening it if you don't wan to merge it? ^^

miniche:
I can open an issue instead of this PR too.
I think that it's easier to talk about this in a pull request: comment code parts, following the work and adding comments between two commits on a "timeline".

miniche:
Currently, Tornado doesn't seems a better solution than CherryPy (tested on a development environment):
- The memory consumptions are equals
- No difference between CPU utilization

Cons:
- Tornado doesn't have any session management system. It supports natively authentification with Twitter, Facebook or Google, but not with a custom one.
- Tornado Templates is inspired by Jinja. However, all variables must be defined in the controller, otherwise Tornado raise a beautiful exception, even in a 'if' bloc. To handling all cases, we must add a "try/catch" bloc in the template or play with the "locale()" function.

I close this Pull Request for now, we can follow the development of both in their next releases.

PR #81, opened 2013-06-25
Removed Flot.js library

miniche:
The flot library (http://www.flotcharts.org/) is not used anymore.
I think we will debate about a new graph library in a next release, when we will refactor the internal web server.

This pull request deletes also an old template file (dashboard_mac.tpl), not used since the version 0.1.0.


miniche:
Hey guys! Ready for your comments and your review ;-)

Ninir:
Here is a list of libraries we should have a look to:
- http://www.chartjs.org/  (even working on IE7/8 with polyfills)
- http://d3js.org/

Ninir:
:+1:

Guimove:
:+1:

PR #84, opened 2013-10-30
Fixed "sudo -n" and "sudo -v" for OS X Mavericks

miniche:
CentralReport must have administrative privileges to install or uninstall itself.
In the bash scripts (install.sh and uninstall.sh), we are testing whether the sudo session still alive. On OS X 10.9 Mavericks, the command "sudo -n" returns 0, even if the sudo session is not alive.

Replaced by "sudo -v -p", with a custom message.
If the sudo session is alive, the message is not displayed.

capture decran 2013-10-30 a 23 21 02

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -48,6 +48,7 @@ Version Alpha 0.3.0 (Not released yet)
 
     - Fix #81: Removed flot.js library and 'dashboard_mac' template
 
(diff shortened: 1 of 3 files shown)


miniche:
The release notes for the 0.3.0 had been updated with this fix:
http://docs.centralreport.net/unix-agent/releases/0.3.0

Issue #85, opened 2013-10-30
OS X 10.9 Mavericks and the memory compression


miniche:
As you know, OS X 10.9 Mavericks introduces big features like "Time coalescing" and "App nap" for the processor consumption.
It introduces the memory compression too (you can read http://www.apple.com/osx/advanced-technologies/ and http://apple.stackexchange.com/questions/107126/what-scale-or-measure-does-mavericks-memory-pressure-adhere-to).

In fact, Apple had created more memory page types.
On OS X 10.8 Mountain Lion, the "vm_stat" command returns something like this:
![capture decran 2013-10-30 a 23 48 53](https://f.cloud.github.com/assets/1638158/1442036/7a9850e8-41b5-11e3-8906-9f14eb6720a6.png)

On OS X 10.9, it's a little more complex:
![capture decran 2013-10-30 a 23 49 37](https://f.cloud.github.com/assets/1638158/1442041/924924d8-41b5-11e3-8657-e36b95e17e02.png)

To define the total memory installed, CentralReport adds these values:
- Pages free
- Pages active
- Page inactive
- Pages resident

With this method, on my Mac with 8 GB RAM installed, when OS X compresses the memory, CentralReport returns values like 5 or 6 GB in my case.

```
"memory_free": "465.38 MB",
    "memory_total": "5.42 GB",
    "memory_used": "4.97 GB",
    "memory_state": "alert"
```

We must look at the new behavior of OS X with the memory and includes another page type in the calculation.

miniche:
Solved by #88. Now available in master.

PR #87, opened 2013-11-06
New host module and fixed memory collector for OS X 10.9 Mavericks

miniche:
Apple had integrated the WKdm algorithm to compress data in the RAM, in OS X 10.9 Mavericks.
Unused memory can be compressed to free more memory, without swapping.

I have tested this behavior on my MacBook Pro with 8 GB of RAM. Mac OS X can give 12 GB of memory to the system and to the applications, before starting swap on the hard drive/SSD! Just incredible.
During this process, the CPU occupation is higher.

You can read more about it on:
- http://www.apple.com/osx/advanced-technologies/
- http://terpconnect.umd.edu/~barua/matt-compress-tr.pdf

So, to do it, Apple had added new page types for the memory (see #85).
CentralReport can now analyse them and displays right data about memory.

I have added the "cr.host" module. It contains all code to determine the current host and to get information about it. It was previously in the "collector" module, which is destined to died in a next release.

### DIFF
--- a/centralreport/centralreport.py
+++ b/centralreport/centralreport.py
@@ -19,6 +19,7 @@ import os
 CR_CURRENT_DIR = os.path.abspath(os.path.dirname(__file__))
 sys.path.insert(0, os.path.join(CR_CURRENT_DIR, 'libs/urwid-1.1.1.zip'))
 
+from cr import host
 from cr import log
 from cr import threads
 from cr.utils import text
@@ -80,6 +81,9 @@ class CentralReport(Daemon):
         CentralReport.starting_date = datetime.datetime.now()  # Starting date
         CentralReport.configuration = Config()  # Getting config object
 
+        # Getting data about the current host...
+        host.get_current_host()
+
         # The log level can be personalized in the config file
         if Config.CR_CONFIG_ENABLE_DEBUG_MODE is False:
             try:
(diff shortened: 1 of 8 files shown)


miniche:
Guys,

This pull request is an important fix to support the new OS X Mavericks. However, I had to change modules and functions which impact all OS.
In this case, we can't release a version "0.3.1", but a new "0.4.0", do you agree?

I will test this new code on Linux this afternoon, and I think it's better to merge this fix in master as soon as possible.

miniche:
Team,

I've tested this improvement on Mac OS X 10.8 and 10.9, Debian 6 and 7, CentOS 6 and Ubuntu 13.04: works well.
So, what do you prefer for the version number?

And a review is welcome :wink:

centralreport/centralreport.py
```diff
@@ -80,6 +81,9 @@ def run(self):
         CentralReport.starting_date = datetime.datetime.now()  # Starting date
         CentralReport.configuration = Config()  # Getting config object
 
+        # Getting data about the current host...
```
Ninir: Not useful comment detected! :smile:

Ninir:
Seems good to me :)

miniche:
Great :)

And about the version number, what do you prefer? (https://github.com/CentralReport/CentralReport/pull/87#issuecomment-27867470)

Ninir:
It's a minor improvement, 0.4.0 is good for me :) ! :+1:

miniche:
This pull request is not valid: this branch must be merged in develop, not in master. Replaced by #88.

Issue #75, opened 2013-05-14
Debian 7 and Vagrant


miniche:
Hey the team!

Debian 7 "Wheezy" is out! (Official link: http://www.debian.org/News/2013/20130504)
We must make some tests with this new Debian version to be sure that CentralReport remains 100% compatible.
I'm downloading it, and I will try to do first tests tonight.

To perform more accurate tests, and with more flexibility, we could look towards Vagrant (http://www.vagrantup.com/). We can launch all tests easier, and make sure that all updates works on all "basic" configurations.

However, we must create the bash and python unit tests before!

miniche:
CentralReport works well on Debian 7! 
The disk is identified by its UUID, and not by its name. I will read the Debian changelog to collect more data about this.

![capture decran 2013-05-17 a 01 11 02](https://f.cloud.github.com/assets/1638158/515633/ff2496f2-be7d-11e2-9287-eb773995c80e.png)

miniche:
The PR #52 adds the support for the disk UUID detection. Disk names will be displayed properly in this case.

miniche:
The Vagrant part will be added with the pull request #77

miniche:
Solved by #92 and #77. Available in the develop branch, soon in master.

PR #92, opened 2014-01-07
Updated disk collector for Linux (get disks by uuid)

miniche:
This pull request improves the compatibility with Debian 7, and solves the issue #75.
It displays the correct name of the disk instead of its UUID.


miniche:
Ready to be reviewed and merged :)

PR #77, opened 2013-06-02
Unit testing and behavior tests

miniche:
The CentralReport project is growing, it's time to implement tests!
Tests allow to ensure the integrity of the application is kept between versions and to ensure that all functionalities works well all the time.

CentralReport have two main parts:
- The bash part (installer/uninstaller and other related tools)
- The Python part (the application)

This pull request creates the basis of unit and system tests. There is a lot of work with the amount of functions and methods. All tests will be implemented while CentralReport is developed.

The following methods are ready:
- [x] Tests for the installer and the uninstaller (system tests)
- [x] Unit tests for the "utils" Python package
- [x] Tests with [Travis-CI](https://travis-ci.org/CentralReport/CentralReport), supporting the tests on multiple Python versions
- [x] Tests with [Vagrant](http://www.vagrantup.com/)
## Start tests

All tests can be run with the `run_tests.sh` script, in the "tests" directory. This script accepts the following options:

``` bash
-a Performs all tests, including the Vagrant tests
-p Only performs the Python unit tests
-s Only performs the system tests (installer/uninstaller)
-v Only use Vagrant to test CentralReport on virtual machines (in this case, Python unit tests and system tests are performed on each VM)
```

tests
## Vagrant

Vagrant is a powerful software to test your project on many virtual machines, with the help of VirtualBox or VMware.
To start, I have set up three virtual machines:
- Debian 6.0.5 x64
- CentOS 6.4 x64
- Ubuntu Server 12.04 x64

They are hosted at http://static.centralreport.net/vagrant/.
If you want to test this feature, keep in mind that the size of each virtual machine is approximately 300 Mb.
It requires the lastest Vagrant version and VirtualBox >= 4.2
## In the future

There are lots of tests to write now. To avoid a long time wasting, they will be written while new features are added.
TODO:
- Unit tests for all bash functions (with shunit2)
- More tests for bash methods (install/uninstall/manager/CLI/...)
- More unit tests for Python
- System tests for Python
## Remains to be done in this pull request
- [ ] Update the NEWS file

PR #94, opened 2014-01-13
Added timezone management in the Javascript part

miniche:
The timezone can be different between the host and the client.

With #77, the datetime to timestamp function now returns an UTC time instead of a local time.
The javascript functions handle an UTC time too to calculate time intervals.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -14,6 +14,8 @@ Version Alpha 0.5.0 (Not released yet)
     - Fix #93: Fixed uptime displayed on the web server
         * Seconds are now refreshed in uptime data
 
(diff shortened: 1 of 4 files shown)


miniche:
Good for me.
Waiting for the Travis build and the @Ninir review :)

miniche:
Travis builds are... OK :D
https://travis-ci.org/CentralReport/CentralReport/builds/16898547

miniche:
Ding dong! @Ninir is here? :smile:

Ninir:
Seems ok :)

